PR #26, opened 2017-04-18
GraphQL schema optimization

rchall-gladly:
Based on this: https://github.com/facebook/relay/issues/112#issuecomment-170648934


rchall-gladly:
No longer useful because we are now updating for Relay Modern

PR #33, opened 2017-04-26
Widgets - Bookmarks widget

rchall-gladly:


graphql/database/widgets/bookmarkWidget.js
```diff
@@ -0,0 +1,183 @@
...
+   */
+  constructor(userId, widgetId, widget) {
+    this.widget = widget;
+    this.userId = userId;
+    this.widgetId = widgetId;
+    this.bookmarks = widget.data.bookmarks;
+
+    /**
+    * Updates the bookmarks of the current UserWidget.
+    * @return {Object<UserWidget>}  Returns the UserWidget with
```
kmjennison: @rchall-gladly I believe the `@return` should be `Promise<>` rather than `Object<>` for this method and all others that call `Async`?

graphql/database/widgets/widgets.js
```diff
@@ -0,0 +1,265 @@
...
+class Widget extends BaseModel {
+
+  /**
+   * Creates a Widget instance.
+   * @param {string} id - The instance id in the database.
+   */
+  constructor(id) {
+    super(id);
+
+    this.name = '';
```
kmjennison: The `name` field on the widgets seems redundant when we have the `type` and we'll localize the name anyway. Remove?

PR #34, opened 2017-04-26
Add a simple search widget with support for Google search only.

rchall-gladly:


dynamodb/fixtures/UserWidgetsData.json
```diff
@@ -27,5 +27,14 @@
...
     	]
     }
+  },
+  {
+    "userId": "45bbefbf-63d1-4d36-931e-212fbe2bc3d9",
+    "widgetId": "a458d2da-550b-414f-895a-7bcb2cbc9d62",
+    "enabled": true,
+    "data": {
+        "engine": "Google",
+        "searchApi": "https://www.google.com/search?q="
```
kmjennison: The `searchApi` functionality makes more sense to keep in the component, because it will be the same for every instance of UserWidgetData.

PR #61, opened 2017-06-22
Add eslint

rchall-gladly:


web/package.json
```diff
@@ -98,11 +99,12 @@
     "deploy": "yarn run build && serverless deploy && serverless syncToS3",
     "relay": "relay-compiler --src ./src/ --schema ./data/schema.graphql",
     "relay:watch": "nodemon --watch src/ --watch data/schema.graphql --ignore **/__generated__/**/* --exec \"yarn run relay\"",
-    "test": "npm-run-all -s test:run test:codecov",
+    "test": "npm run test:lint && npm-run-all -s test:run test:codecov",
```
kmjennison: For all package.json files, include linting as one of the `npm-run-all` commands. `npm-run-all -s` is just a cross-platform replacement for using `&&` between scripts.

web/package.json
```diff
@@ -98,11 +99,12 @@
...
     "test:run": "jest --env=jsdom --coverage",
     "test:codecov": "if-env NODE_ENV=test && codecov || echo 'Skipping test coverage logging.'",
     "test:watch": "jest --env=jsdom --watch",
-    "build:langs": "node scripts/translate.js"
+    "build:langs": "node scripts/translate.js",
+    "test:lint": "standard --fix"
```
kmjennison: Let's just run `standard` without the `--fix` flag, because we don't expect our test code to modify the source.

PR #119, opened 2017-11-29
Fix donate hearts

kmjennison:
Update DynamoDB schema to match expected schema.


kmjennison:
Merging into master to delete/recreate table on production.

PR #125, opened 2017-11-30
Add to referral data

kmjennison:
* Change hash key name for ReferralDataLog table
* Add a `referralChannel` field to the referral data object
* Add some tests for getting referral data from local storage


kmjennison:
Going to merge into master to delete the table on production, then recreate it with a new hash key name in another PR.

PR #127, opened 2017-11-30
Delete VcDonationLog table (to recreate with a new key schema)

kmjennison:


kmjennison:
Going to merge into master to delete prod table, then recreate.

PR #154, opened 2017-12-05
Format MomentJS for 24 Hour Time

### LINKED ISSUE #152: 24-hour clock should show 00:00 for midnight, not 24:00
bad midnight clock

pshah123:
Previously the Clock widget used `"k:mm"` to format 24 hour time but the correct format-code is `H`, not `k`. Changed moment format to `"H:mm"` for 24 hour time (starts at 0).

Fixes #152


kmjennison:
Thanks! This should help people not lose track of time tonight :)

PR #236, opened 2018-03-23
Use AutoDll Webpack plugin

kmjennison:
* Use [autodll-webpack-plugin](https://github.com/asfktz/autodll-webpack-plugin) for vendor module bundling instead of CommonChunksPlugin
* Reorganized web app's `dependencies` to include only the packages that are bundled into client JS
* Confirmed filename hashes change when expected:
  * Changing app JS code changes the filename hash for app.js but *not* the filename hash for ads.js or vendor.js
  * Updating a vendor module changes the filename hash for both vendor.js and app.js


kmjennison:
Bug: the vendor file hash can change without updating the file hash, which breaks the page.

To reproduce:
1) Build the web JS and note the chunkhashes for app.js and vendor.js.
2) Add a comment to vendor.js.
3) Rebuild the web JS. The chunkhash fro vendor.js (but not app.js) should have changed, and the app is broken because of an old vendor variable reference in app.js code.

kmjennison:
In fact, could not reproduce broken hash build. Going to deploy as-is. To make sure we catch any future errors, I added simple acceptance testing in #237.

kmjennison:
Dev deployment includes the development version of React. Need to make sure vendor bundle is built with production settings.

Issue #202, opened 2018-01-26
Set daily ad views limit and "Hearts from tabs" limit


kmjennison:
To prevent abuse by bots and tab spamming.

PR #255, opened 2018-04-19
Max daily Hearts from opening tabs

kmjennison:
Set a maximum of 150 Hearts that a user can earn from opening tabs within a single UTC day.

* Analytics change: log all tabs, even those that we consider "invalid" and do not earn any Hearts
* Add "tabsToday" field to user model
* Adjust BaseModel class to set all field defaults before running deserializers
* Show a message to users when they reach maximum Hearts so they understand why their Hearts aren't increasing

### DIFF
--- a/graphql/data/schema.js
+++ b/graphql/data/schema.js
@@ -217,6 +217,10 @@ const userType = new GraphQLObjectType({
       type: GraphQLInt,
       description: 'User\'s all time tab count'
     },
+    tabsToday: {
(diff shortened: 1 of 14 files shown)


kmjennison:
New tab when max Hearts reached (add checkmark to Heart):

max-hearts-reached

On hover:

max-hearts-reached-hover

kmjennison:
Closes #202.

PR #309, opened 2018-08-17
Handle missing user

kmjennison:
Handle the case when a user is authenticated but does not have a user on the server side. This may happen if Firebase authentication succeeds but the request to our server fails (e.g., network error, or user navigates away). Currently, a user would be stuck in this broken state. Now, we try to create the user on the server side and refetch data (when on the dashboard view).

* Add a "code" field to our returned GraphQL error
* Throw a custom UserDoesNotExist error when querying for a User item that does not exist
* Don't log UserDoesNotExist errors
* Move the logic to create a new user into its own module
* Add some missing tests

### DIFF
--- a/graphql/utils/exceptions.js
+++ b/graphql/utils/exceptions.js
@@ -12,7 +12,7 @@ class ExtendableError extends Error {
 
 class NotImplementedException extends ExtendableError {
   constructor () {
-    super('Required method is not implmented.')
+    super('Required method is not implemented.')
   }
 }
 
@@ -24,7 +24,7 @@ class UnauthorizedQueryException extends ExtendableError {
 
 class UserReachedMaxLevelException extends ExtendableError {
   constructor () {
-    super('There are no more levels user at max level.')
+    super('There are no more levels. The user is at the max level.')
   }
 }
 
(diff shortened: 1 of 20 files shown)


graphql/utils/exceptions.js
```diff
@@ -10,47 +10,62 @@ class ExtendableError extends Error {
   }
 }
 
+export const METHOD_NOT_IMPLEMENTED = 'METHOD_NOT_IMPLEMENTED'
 class NotImplementedException extends ExtendableError {
   constructor () {
-    super('Required method is not implmented.')
+    super('Required method is not implemented.')
+    this.code = UNAUTHORIZED_QUERY
```
kmjennison: Fix this code

web/src/js/authentication/helpers.js
```diff
@@ -44,6 +48,20 @@ const allowAnonymousUser = () => {
...
+  // We may want to handle this in QueryRenderers:
+  // - In GraphQL, create a custom UserDoesNotExist error
+  // - UserModel extends BaseModel's `get` method; if the item
+  //   does not exist, throw UserDoesNotExist
+  // - In formatError, return error codes. Also filter out some
+  //   errors that we do not want to log because they're sometimes
+  //   expected, like UserDoesNotExist errors.
+  // - In QueryRenderers, if the error indicates that the user does
+  //   not exist, create the user and then re-query.
+
```
kmjennison: Remove

web/src/js/components/Dashboard/__tests__/DashboardView.test.js
```diff
@@ -0,0 +1,232 @@
...
+          ],
+          operation: { foo: 'bar' },
+          variables: { foo: 'baz' }
+        }
+      },
+      props: null,
+      retry: mockRetryFn
+    })
+
+    // Failed to createa  new user.
```
kmjennison: Fix typo

PR #314, opened 2018-09-01
Anonymous user support [first piece]

kmjennison:
This is the first step toward supporting anonymous users. This merge should not have any effect on the end-user experience, and the anonymous user functionality is disabled via feature flag.

In the merge, we:
* Allow users to not have an email address in our authorization logic
* Allow users in our database to not have an email address
* Move the client-side logic for authentication redirects and user creation into its own module
* Incorporate #309, which creates a new user in our database when one does not exist (in the DashboardView and AuthenticationView), which is a more common scenario with anonymous users
* For all users, immediately create new anonymous user accounts (in Firebase authentication and our database) when a new user visits the page. This gives us parity in comparing the conversion rates between anonymous users and users who are required to sign in.
* Assign all new users to a test group (stored on the client) for anonymous sign-in
* Add logic for whether a user is required to sign in. An anonymous user is _not_ required to sign in if all of these are true:
  * They are in the "anonymous" experimental test group
  * They joined less than 2 days ago. In other words, users are required to sign in after two days, which prevents them from losing access to their tab page data.
  * The feature flag (`FEATURE_FLAG_ANON_USER_SIGN_IN`) is enabled. It's set to `false` in this merge.
* On sign-in, upgrade anonymous users to signed-in and store their email address. Here. we handle merge conflicts—when an anonymous user signs in as an already-existing user—by deleting the anonymous user in Firebase and marking the anonymous user in our database as "merged into an existing user".

The main changes that remain (not addressed in this PR) are:
* Only reward user referrers after the referrer verifies their email after sign-in
* Hide the logout button for anonymous users (because logging out will likely lose their data)
* Handle missing username and email in the referral URL, stats, and account page
* Show a "sign in" call to action on the dashboard for anonymous users
* On the sign-in page, add a message specific to anonymous users to tell them why we’re requiring sign-in
* Log the user's experimental group
* On log out, delete any local storage data we don't want to keep


web/src/js/authentication/__tests__/helpers.test.js
```diff
@@ -0,0 +1,196 @@
...
+    expect.assertions(2)
+
+    const checkAuthStateAndRedirectIfNeeded = require('../helpers')
+      .checkAuthStateAndRedirectIfNeeded
+    const user = {
+      id: 'abc123',
+      email: null,
+      username: 'foo',
+      isAnonymous: false,
+      emailVerified: true
```
kmjennison: Change emailVerified to `false`

PR #318, opened 2018-09-13
Reward referrers when their referred users verify their email address

kmjennison:
This PR aims to reduce potential user referral fraud (to earn hearts), which is much easier when we allow anonymous user creation. Because some code logic is a little tenuous, this PR also increases the likelihood that we'll fail to reward a referrer.

The code here is a little messy, because Firebase unfortunately [does not support](https://stackoverflow.com/q/43503377) triggering a cloud function when an email is verified, nor is there an "onEmailVerified" callback from the JS SDK. We deal with this limitation by:
1. setting a user's email as verified on sign-in if the Firebase claims show the email is verified. The email address will be immediately verified for identity providers that are auto-verified (such as Google sign-in).
2. attempting to log a user's email as verified when they visit an `/newtab/auth/action/` URL (the page they'll visit when they click an email verification link)
    * Here, we poll Firebase a number of times to see if the email is verified.
    * If the email's verified, we force-reload the user ID token (to refresh the email verification claim) and log the email as verified.
    * This is imperfect. It will fail if: 1) Firebase takes more than a few seconds to say the email is verified (because we'll give up polling); or 2) the user navigates away before we're able to confirm and log the email verification.

When a user's email is logged as verified, we'll reward the referrer if they haven't already been rewarded. Logging email verification can and likely will occur multiple times (each time a user signs in). That's ok.

This PR also adds two test util functions to aid in async web testing: `flushAllPromises` and `runAsyncTimerLoops`.

### DIFF
--- a/graphql/data/schema.js
+++ b/graphql/data/schema.js
@@ -43,6 +43,7 @@ import {
 import UserModel from '../database/users/UserModel'
 import createUser from '../database/users/createUser'
 import setUsername from '../database/users/setUsername'
+import logEmailVerified from '../database/users/logEmailVerified'
 import logTab from '../database/users/logTab'
 import logRevenue from '../database/userRevenue/logRevenue'
 import logUserDataConsent from '../database/userDataConsent/logUserDataConsent'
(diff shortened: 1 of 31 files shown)


kmjennison:
Fix: this will currently re-reward past user referrers when a user logs in again and we log their email as verified for the first time.

kmjennison:
Fixed: prevent re-rewarding past user referrers.

FIXME: the client calls "create user" with the old (anonymous) ID token when signing in with Google, which should show a verified email. Need to refresh the ID token before creating the user.

kmjennison:
Fixed: we refresh the ID token before creating the user.

PR #321, opened 2018-09-18
Log the user's test groups on user creation

kmjennison:
Some tests, like the anonymous user sign0n, assign the user to test groups on the client side before user creation. This PR makes sure we log the user's test group for the anonymous user sign-in when we create the user.

* Add a numeric `testGroupAnonSignIn` field to the User model
* Add an `ExperimentGroupsType` GraphQL object
* Send an "experiment groups" object in the mutation when creating a new user, then store the user's experiment groups in the database

### DIFF
--- a/graphql/data/schema.js
+++ b/graphql/data/schema.js
@@ -313,9 +313,6 @@ const userType = new GraphQLObjectType({
       type: GraphQLBoolean,
       description: 'Whether this user was created by an existing user and then merged into the existing user'
     }
-    // experimentGroups: {
-    //   type: ExperimentGroupsType,
-    // }
   }),
   interfaces: [nodeInterface]
 })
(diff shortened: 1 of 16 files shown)


graphql/data/schema.js
```diff
@@ -291,6 +313,9 @@ const userType = new GraphQLObjectType({
       type: GraphQLBoolean,
       description: 'Whether this user was created by an existing user and then merged into the existing user'
     }
+    // experimentGroups: {
+    //   type: ExperimentGroupsType,
+    // }
```
kmjennison: Remove

graphql/utils/__tests__/experiments.test.js
```diff
@@ -0,0 +1,93 @@
...
+      anImportantTest: 'this is invalid!'
+    })
+    expect(expGroups).toEqual({
+      anonSignIn: null,
+      fooTest: 1,
+      anImportantTest: null
+    })
+  })
+
+  test('returns all null experiment groups when nothing is provdied', () => {
```
kmjennison: Fix typo

PR #325, opened 2018-09-20
Safer database item creation and update

### LINKED ISSUE #322: Make updates fail if the database item does not exist; make creates fail if the item does exist
It's possible to try to update an item that doesn't exist; e.g. if there's a race condition, or if the original item creation request failed. In that case, the item will be missing the required `created` field and perhaps other fields, which will cause bugs. We've seen this in a few instances for user items.

It's also possible to try to create an item when one already exists. Currently, we'll overwrite the item, which is dangerous.

kmjennison:
Fixes #322 

This PR prevents updates on non-existing database items (prevent items with missing fields) and prevents creations on existing items by default (prevent data loss by overwrite).

### DIFF
--- a/graphql/database/base/BaseModel.js
+++ b/graphql/database/base/BaseModel.js
@@ -395,17 +395,22 @@ class BaseModel {
   static async update (userContext, item, params = {}) {
     // logger.debug(`Updating item in ${this.tableName}: ${JSON.stringify(item, null, 2)}`)
     const self = this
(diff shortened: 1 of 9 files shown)


kmjennison:
FIXME: item creation may be failing the conditional check. Investigate.

kmjennison:
Fixed: be sure to explicitly create UserWidget items rather than rely on update.

Issue #311, opened 2018-08-28
Switch to Prettier code formatting


kmjennison:
Possibly add a pre-Git-push handler to format code.

PR #352, opened 2018-09-29
add prettier with pre-commit hook

kristinn93:
As suggested by #311 this PR is setting up prettier with pre-commit hook using [husky](https://github.com/typicode/husky)


kmjennison:
Thanks! Holding off on merging this for now because I'm doing some other refactoring that will conflict.

kmjennison:
Closed in #427. Thanks!

PR #584, opened 2019-05-24
Set restrictive stack policies on all stacks

### LINKED ISSUE #582: Set safer resource deletion, updating, and retention policies
Currently, it's too easy to accidentally delete a stack or resource; for example, by deploying a stack with the same ID as an existing stack, which is what I did yesterday to cause site downtime. This is particularly worrisome for resources like DynamoDB tables.

We should make our stacks more conservative at accepting updates and deletion.

- [x] Retain some resources, such as DynamoDB tables, even when the stack is deleted
  * See [CF's DeletionPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html)
  * To test:
    1. Set the deletion policy to "retain" for a non-critical DDB table; deploy to dev
    2. In our stack config, delete another non-critical DDB table; deploy to dev and confirm it's removed
    3. In our stack config, delete the retained non-critical DDB table; deploy to dev and confirm the table is retained
    4. Manually delete the retained table and re-deploy to dev to recreate the two tables
  * Set the deletion policy to "retain" for all S3 buckets and DynamoDB tables
- [x] Define a stack policy that, by default, prevents modifying any stack resources
  * See [CF docs on stack policies](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/protect-stack-resources.html)
  * Define this in the [`stackPolicy` key in serverless.yml](https://serverless.com/framework/docs/providers/aws/guide/serverless.yml/)
  * To test:
    1. Set a completely restrictive stack policy on a stack
    2. In our stack config, delete a non-critical resource; deploy to dev
    3. Confirm that the deploy fails because deleting the resource is not allowed
    4. Remove the stack config and re-deploy; confirm that deleting the resource still fails, because it should be impossible to remove an existing stack config
- [x] Add termination protection to our stacks
  * See [CF docs on termination protection](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-protect-stacks.html)
  * This is unfortunately [not yet supported by Serverless](https://github.com/serverless/serverless/issues/4558)
  * Let's just manually set termination protection in the AWS console
- [x] Add an UpdateReplacePolicy set to "Retain" where needed
  * See: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatereplacepolicy.html

kmjennison:
Addresses part of #582 

In prior commits, I confirmed:
* Deleting a DynamoDB table fails with this stack policy
* Deleting the stack policy from serverless.yml does not remove the stack policy from Cloudformation

This is a multi-PR deployment.

**First deployment (this PR):**
- [x] set all stack policies to `Allow` and deploy to dev
- [x] set deployment to be forced with `serverless deploy --force`
- [x] confirm all stack policies are set in Cloudformation and services work as expected
- [ ] deploy to production and  confirm all stack policies are set in Cloudformation

**Second deployment:**
- [ ] make sure deployment is forced with `serverless deploy --force`
- [ ] set all stack policies to `Deny` (excepting `Update:Modify` on lambda functions) and deploy to dev
- [ ] deploy again to make sure future deploys will succeed
- [ ] deploy to production and check all stack policies are set as expected

**Third deployment:**
- [ ] disabled forced deployment by removing `--force` from `serverless deploy --force`

### DIFF
--- a/dynamodb/package.json
+++ b/dynamodb/package.json
@@ -41,7 +41,7 @@
     "format": "prettier \"**/*.{js,jsx,ts,tsx,json,css,scss,md}\"",
     "format:fix": "yarn run format --write",
     "lint": "eslint \"**/*.{js,jsx}\"",
-    "deploy": "serverless deploy -v"
+    "deploy": "serverless deploy -v --force"
   },
   "repository": {
     "type": "git",
--- a/dynamodb/package.json
+++ b/dynamodb/package.json
@@ -41,7 +41,7 @@
     "format": "prettier \"**/*.{js,jsx,ts,tsx,json,css,scss,md}\"",
     "format:fix": "yarn run format --write",
     "lint": "eslint \"**/*.{js,jsx}\"",
-    "deploy": "serverless deploy -v --force"
+    "deploy": "serverless deploy -v"
   },
   "repository": {
     "type": "git",
--- a/dynamodb/serverless.yml
+++ b/dynamodb/serverless.yml
@@ -7,14 +7,14 @@ provider:
   runtime: nodejs8.10
   region: us-west-2
   stage: ${env:SLS_STAGE, self:custom.stageDefault}
-  # Let's be conservative and prevent any unexpected changes
-  # to the stack. See:
-  # https://github.com/gladly-team/tab/issues/582
-  stackPolicy:
-    - Effect: Deny
-      Principal: "*"
-      Action: "Update:*"
-      Resource: "*"
+  # # Let's be conservative and prevent any unexpected changes
+  # # to the stack. See:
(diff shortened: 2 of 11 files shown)


kmjennison:
Not all Cloudformation policies are updating as expected. Let's try this:
1. set all stack policies to `Allow` and deploy to dev
2. then, set all serverless deploys to force-deploy and deploy to dev
3. confirm all stack policies are set in Cloudformation and services work as expected
4. set all stack policies to `Deny` and deploy to dev
5. deploy again with minor code changes to make sure the deploy will succeed
6. deploy to production and check all stack policies are set as expected

PR #628, opened 2019-08-09
Make the auth flow support Search branding

kmjennison:
This PR brands our existing auth flow for either Search for a Cause or Tab for a Cause depending on whether the `app` URL parameter value is "search" or "tab". The branding changes the Material UI theme, certain copy, the logo, page favicon, and page title. By default, show Tab for a Cause.

* Add navigation helpers to preserve the URL search string when navigating and to construct a link URL with mondified search params.
* Preserve URL search params in the auth flow when redirecting to the appropriate auth step. This lets us preserve the `app` URL parameter value.
* Pass the `app` URL parameter value to the Firebase "continueUrl" so that the app branding is preserved after email confirmation.
* Remove favicon `link` elements and `title` elements from HTML. Instead, set these in Javascript using `react-helmet`.
* Upgrade `react-helmet` to v6.0.0-beta.
* In root.js, create a workaround for a [`react-helmet` bug](https://github.com/nfl/react-helmet/issues/430) to allow us to modify the favicon.
* Use new Material UI components for the auth pages.
* Support app-specific branding in the ErrorBoundary and FullPageLoader components

FIXME before merging: social sign-in always goes to `/newtab/`. It should go to the appropriate app destination.

Remaining TODOs:
* Use different values for `tosUrl` and `privacyPolicyUrl` for each app in FirebaseAuthenticationUI.js.
* For convenience, redirect `/search/auth/` to `/newtab/auth/?app=search`


kmjennison:
Fixed: successful sign-ins with Facebook and Google redirect the appropriate app, either Tab or Search.

PR #649, opened 2019-09-09
Don't set search state that's derived from the `location` prop

kmjennison:
We've been keeping the search query, search result page number, and search source in state in the SearchPageComponent when it's actually state derived from the `location.search` URL parameter values. It adds complexity and might be causing bugs. Remove these from state altogether.


kmjennison:
Fix a bug: when loading a search result set to a page other than 1, the pagination buttons are incorrect.

PR #656, opened 2019-09-12
Add an "account creation" analytics event for Search

kmjennison:
* Log a "search account creation" analytics event when the user's total searches equals exactly one
* Remove unneeded Quora analytics
* Move Facebook analytics JS to the head
* When pre-rendering, don't run analytics JS that we use on the search app, because that can cause errors in the analytics
* Wrap some analytics JS in try/catch
* Fix an unrelated UI error in async-changing button color for the search intro


kmjennison:
FIXME: we need to be smarter about only firing the account creation event once per user. Right now, it could fire twice.

PR #835, opened 2020-10-03
Open bookmarks in new tab with middle mouse button

### LINKED ISSUE #622: Allow bookmarks to be opened in a new tab with the middle mouse button
I am very much used to mousewheel click links to open them in a new tab. Would be awesome if that was possible with the bookmarks as well.

jeffslofish:
Closes #622


kmjennison (changes requested):
Thanks for the PR. Rather than use mouse events here, I think the better approach here is to simply render anchor tags and let the browser handle click behavior normally. If you make that change, I'd also welcome tests for this (currently untested) component.

jeffslofish:
I messed this up with the merge. I'm going to do a new pull request with just the new changes.

PR #839, opened 2020-10-06
Add option to show seconds on clock

### LINKED ISSUE #805: Option to add seconds to the clock
User requested feature.

jeffslofish:
Closes #805

### DIFF
--- a/dynamodb/fixtures/UserWidgetsData.json
+++ b/dynamodb/fixtures/UserWidgetsData.json
@@ -67,7 +67,8 @@
     "data": {},
     "widgetId": "8b5e572b-7f44-45ea-965b-55e6a22ca190",
     "config": {
-      "format24": false
+      "format24": false,
+      "showSeconds": false
     }
   }
 ]
--- a/dynamodb/fixtures/WidgetsData.json
+++ b/dynamodb/fixtures/WidgetsData.json
@@ -38,6 +38,12 @@
         "type": "boolean",
         "display": "Use 24 hours format",
         "defaultValue": false
+      },
+      {
+        "field": "showSeconds",
+        "type": "boolean",
+        "display": "Show seconds",
+        "defaultValue": false
       }
     ],
     "created": "2017-07-18T20:45:53Z",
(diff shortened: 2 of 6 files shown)


jeffslofish:
Updated tests coming...

kmjennison:
@jeffslofish This looks great! Thanks for the contribution and especially for the associated tests.

The only requested change is for UI: I'd love to add a little more spacing between the settings options. In WidgetConfigComponent [here](https://github.com/gladly-team/tab/blob/1c884b6fa2cb62a0267ae4f48cea12d909227a62/web/src/js/components/Settings/Widgets/WidgetConfigComponent.js#L76), could you please add `marginTop: 8, marginBottom: 8,`?

Before:
before

After:
after

jeffslofish:
I believe there are a couple problems with that:
1. I think you meant to put the css on BooleanWidgetConfig [here](https://github.com/gladly-team/tab/blob/1c884b6fa2cb62a0267ae4f48cea12d909227a62/web/src/js/components/Settings/Widgets/WidgetConfigComponent.js#L49), not on the ChoicesWidgetConfig which controls the search engines.

2. Assuming you meant to put the css on the BooleanWidgetConfig, then putting 8px of margin on the top and bottom of each element, will push the top setting down so that the space from "Clock" to "Use 24 hours format" is 8px more than the space from "Search" to "Search engine", which IMO makes it look a little wonky. I suggest just using `marginBottom: 16,`. (And ideally the last Boolean element would not have the marginBottom, but I don't see an easy way to accomplish that with the way the inline styles are currently set up.) I have shown the two versions in each of these images below:

Screen Shot 2020-10-06 at 4 05 27 PM
Screen Shot 2020-10-06 at 4 05 56 PM

I can obviously do it any way, but I think the option with just using `marginBottom: 16,` looks better. What do you think? Am I overcomplicating this, lol?

kmjennison:
Yep, sorry for the incorrect line link! (And for these messy components—they're slowly making their way toward a refactor 😄 )

I agree that your layout suggestion looks better 👍

kmjennison:
@jeffslofish This is now live on production! 🎊  Thanks for the contribution!

PR #863, opened 2021-01-08
chore(web): fix prettier config

iamclaytonray:


kmjennison (changes requested):
Thanks! In fact, all of our `.prettierrc` files suffer from this error (top-level and in `dynamodb`, `graphql`, `lambda`, and `redis` subdirectories). Let's update those too and also change all to `.prettierrc.json` to be explicit about syntax.

web/.prettierrc
```diff
@@ -5,5 +5,5 @@
   "semi": false,
   "singleQuote": true,
   "tabWidth": 2,
-  "trailingComma": es5,
+  "trailingComma": "es5",
```
kmjennison, line 8: Remove trailing comma here, as JSON doesn't support it.

iamclaytonray: JSON does not, no, but unless you only format JSON, it'd still be relevant for JS/TS. But I can remove it!

PR #866, opened 2021-01-25
Web start script redirects to newtab or search

Sandalf:
I noticed that when you do `yarn run start:newtab` or `yarn run start:search` the browser opens on the root path and not on  `/newtab/` or `/search/`, so I added a redirect for that.


kmjennison:
Thanks for the contribution!

PR #875, opened 2021-02-17
Filter background images by category when fetching them

kmjennison:
This ensures the background images page shows only those from the "legacy" category.


amaliwanag2 (approved):
LGTM 🧀

PR #878, opened 2021-03-03
update UserImpactModel default values

amaliwanag2:


amaliwanag2:
@kmjennison by default we want a new user to be 1 visit away from achieving their first treat

kmjennison (approved):

PR #881, opened 2021-03-12
add setHasViewedIntroFlow and mutation

amaliwanag2:


graphql/data/schema.js
```diff
@@ -480,6 +481,10 @@ const userType = new GraphQLObjectType({
       type: GraphQLBoolean,
       description: 'If true, serve the new Tab V4 app.',
     },
+    hasViewedIntroFlow: {
+      type: GraphQLBoolean,
```
kmjennison, line 485: Make non-null.

...as I should have for some other fields here :)

kmjennison (changes requested):

kmjennison (approved):
LGTM 🍝

PR #885, opened 2021-03-20
TFAC-285 Set cats background for new Tab for Cats user

jedtan:


graphql/database/users/createUser.js
```diff
@@ -51,6 +55,14 @@ const createUser = async (
       ? {
           extensionInstallTimeApprox,
         }
+      : null,
+    v4BetaEnabled
+      ? {
+          backgroundImage: await getRandomBackgroundImage(
```
kmjennison, line 61: I believe using `await` in a property value will cause assigning `undefined` to `userInfo` here, which is synchronous. That's probably the cause of the failing tests.

kmjennison, line 61: Like you suggested, you will have to map the background image to modify it a little. [Here's an example](https://github.com/gladly-team/tab/blob/master/graphql/database/users/setBackgroundImageDaily.js#L22) where we do that elsewhere. The timestamp field is when the background was last set, so you can use a current ISO timestamp.

kmjennison (changes requested):
This is looking really great! Nice job.

graphql/database/users/createUser.js
```diff
@@ -53,6 +57,24 @@ const createUser = async (
...
+    const backgroundCatImage = await getRandomBackgroundImage(
+      userContext,
+      BACKGROUND_IMAGE_CAT_CATEGORY
+    )
+    userInfo = {
+      ...userInfo,
+      backgroundImage: {
+        id: backgroundCatImage.id,
+        image: backgroundCatImage.image,
+        thumbnail: 'dummy',
```
kmjennison, line 72: Sorry, I see [the thumbnail field is set to required in UserModel](https://github.com/gladly-team/tab/blob/master/graphql/database/users/UserModel.js#L168), but it actually no longer is. Go ahead and make it optional there and remove this thumbnail placeholder altogether.

kmjennison (changes requested):
See single comment + it looks like unit tests are still failing.

kmjennison (changes requested):
We need to update the schema here. I'll jump in to update this.

Getting this error when testing a new user signup on the dev site:
```
RelayNetwork: No data returned for operation `CreateNewUserMutation`, got error(s):
Variable "$input" got invalid value { userId: "[redacted]", email: null, referralData: {}, experimentGroups: { searchIntro: "NONE", referralNotification: "NONE" }, extensionInstallId: "[redacted]", extensionInstallTimeApprox: "2021-03-25T18:47:50.446Z", v4BetaEnabled: true }; Field "v4BetaEnabled" is not defined by type CreateNewUserInput.
```

kmjennison (approved):
LGTM! 🍝

PR #887, opened 2021-03-22
seed more cats background images

amaliwanag2:
added some more cat background images to screen grab for the home page.  Figured I might as well add them to the repo for local dev


kmjennison (approved):
LGTM 🍝

PR #888, opened 2021-03-22
fix updating impact from referrer

amaliwanag2:
I actually think we don't want to use getOrCreate here.  If we do, a user might refer another user on tab legacy and we'd end up creating and giving them referral bonuses on our v4 UserImpact document.  If a user does refer someone to Tab for Cats, then their UserImpact record must already be created so I don't think it's an issue

### DIFF
--- a/graphql/database/users/rewardReferringUser.js
+++ b/graphql/database/users/rewardReferringUser.js
@@ -92,7 +92,7 @@ const rewardReferringUser = async (userContext, userId) => {
       pendingUserReferralImpact,
       pendingUserReferralCount,
     } = userImpactRecord
-    await UserImpactModel.update(userContext, {
+    await UserImpactModel.update(override, {
       ...userImpactRecord,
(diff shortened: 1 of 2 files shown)


graphql/database/users/rewardReferringUser.js
```diff
@@ -92,7 +92,7 @@ const rewardReferringUser = async (userContext, userId) => {
       pendingUserReferralImpact,
       pendingUserReferralCount,
     } = userImpactRecord
-    await UserImpactModel.update(userContext, {
+    await UserImpactModel.update(override, {
```
amaliwanag2, line 95: daaaang I'm pretty confident this is the problem.  It would be the referrer's context not the referree's contenxt

kmjennison: Ahh, nice catch!

kmjennison (approved):
LGTM! 🍝

PR #889, opened 2021-03-22
live update graphql on changes

amaliwanag2:
@kmjennison I originally tried using nodemon in the graphql package.json but it wasn't working so I added babel-watch.  Docs say you shouldn't use babel-node or babel-watch in prod, and as far as I can tell the builds mean we don't, but can you confirm that we don't run that start command in prod?


kmjennison (approved):
LGTM 🍝

PR #891, opened 2021-03-23
Send all /newtab/auth* traffic to the legacy Tab app

kmjennison:
This PR adds CloudFront Lambda@Edge logic to route all auth-related requests to the legacy Tab app. It will route:
* any page (no file type suffix) with a path beginning with `/newtab/auth`
* any non-page resource (**must** have a file type suffix) and `Referer` with a path beginning with `/newtab/auth`

This is a workaround until Tab v4 has a fully-implemented auth flow.

Outstanding: we should make sure Tab v4 refreshes upon logout rather than navigating within the app router so that we'll show the Tab legacy auth page.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -241,7 +241,7 @@ before_install:
   # Gulp (used in Prebid.js build)
   - yarn global add gulp
   # AWS CLI
-  - pip install --user awscli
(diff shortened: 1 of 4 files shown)


amaliwanag2 (approved):
:cheese: lgtm

PR #892, opened 2021-03-24
Alec/tfac 206

amaliwanag2:


web/public/index.html
```diff
@@ -381,20 +381,22 @@
       href="https://fonts.googleapis.com/css?family=Roboto:300,400,500"
     />
     <% } %>
-
-    <% if (process.env.REACT_APP_WHICH_APP === "newtab"){ %>
```
kmjennison, line 385: Please wrap the analytics tag in this `if` statement. We exclude JS analytics from the search app, which is built with this same source code.

kmjennison (changes requested):

kmjennison (approved):

PR #897, opened 2021-04-02
TFAC-356 Adding auth_time to Claims List

jedtan:


lambda/src/firebase-authorizer/firebase-authorizer.js
```diff
@@ -98,6 +100,7 @@ function checkUserAuthorization(event, context, callback) {
             uid: decodedToken.uid,
             email: decodedToken.email || null,
             email_verified: decodedToken.email_verified || false,
+            auth_time: decodedToken.auth_time || 0,
           }
```
jedtan, line 105: Should I be setting a custom claim here for future use if they've been used recently? e.g. `admin.auth().setCustomUserClaims(...)`

kmjennison: No, what you did in this PR looks great as-is.

There are two different "claims" here: those that the Firebase token carries, which include these standard claims like `auth_time` that are always available for a verified user token; and the [context we pass to the next Lambda function](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html), which I tend to call "claims" because we're passing on the reliable claims from the token.

kmjennison (approved):
LGTM 🍝

PR #898, opened 2021-04-04
TFAC-329 Adding Reauthentication Logic with Redirect

jedtan:
I think this should do what is desired, since the onAuthProcessCompleted callback sets the redirect URL for the page.

### DIFF
--- a/web/src/js/components/Authentication/Authentication.js
+++ b/web/src/js/components/Authentication/Authentication.js
@@ -32,7 +32,11 @@ import { externalRedirect } from 'js/navigation/utils'
 import Logo from 'js/components/Logo/Logo'
 import searchFavicon from 'js/assets/logos/search-favicon.png'
 import tabFavicon from 'js/assets/logos/favicon.ico'
-import { getUrlParameters, parseUrlSearchString, validateAppName } from 'js/utils/utils'
+import {
+  getUrlParameters,
+  parseUrlSearchString,
+  validateAppName,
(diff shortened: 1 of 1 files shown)


web/src/js/components/Authentication/Authentication.js
```diff
@@ -228,6 +228,11 @@ class Authentication extends React.Component {
           // authentication.
           // https://github.com/gladly-team/tab/issues/589
           this.props.fetchUser()
+
+          const urlParams = getUrlParameters()
+          if (urlParams.reauth === 'true') {
+            await this.onAuthProcessCompleted()
```
jedtan, line 234: Would like thoughts if I should decompose `onAuthProcessCompleted` into just a `redirect` function.

kmjennison: @jedtan Could you explain what you mean here? If you'd just like to move it from a method into its own function—sure!

jedtan: That question makes no sense when i reread it haha. I actually think my original question is moot - i was just thinking of splitting out a separate redirect function that could be called from both `onAuthProcessCompleted` and here (replacing this `onAuthProcessCompleted` call). But now that I think about it, better to reuse the same generic "do this after auth is done" logic here.

kmjennison (approved):
I manually tested, and it works great. LGTM! 🍝

PR #902, opened 2021-04-09
Altruistic giving backend

amaliwanag2:


graphql/database/userImpact/__mocks__/UserImpactModel.js
```diff
@@ -0,0 +1,43 @@
+/* eslint-env jest */
```
amaliwanag2, line 1: this is like a very very primitive mock model.  I started toying with a base model mock class but decided to go as simple as possible for the scope of this work then expand later

kmjennison (approved):
LGTM 🍝

kmjennison:
Do not merge. Waiting on frontend messaging changes.

PR #903, opened 2021-04-10
TFAC-62 Allow Users to Change Email

jedtan:


graphql/database/base/BaseModel.js
```diff
@@ -234,7 +234,9 @@ class BaseModel {
     if (rangeKey) {
       keys.push(rangeKey)
     }
-    // logger.debug(`Getting obj with hashKey ${hashKey} from table ${this.tableName}.`)
+    logger.debug(
```
amaliwanag2, line 237: I think we want to comment this out again

web/src/js/authentication/user.js
```diff
@@ -330,3 +335,25 @@ export const reloadUser = async () => {
...
   await user.reload()
 }
+
+export const updateUserEmail = async newEmail => {
+  const continueUrl = constructUrl(
+    accountURL,
+    { verified: true },
+    { absolute: true }
+  )
+  console.log('continueUrl is ' + continueUrl)
```
amaliwanag2, line 345: X

web/src/js/components/Authentication/EnterEmailForm.js
```diff
@@ -0,0 +1,217 @@
...
+      inputEmail: null,
+    }
+  }
+
+  verifyEmailSent() {
+    this.setState({ verifyEmailSent: true })
+  }
+
+  updateUserFailure(error) {
+    console.log(error)
```
amaliwanag2, line 69: x

web/src/js/components/Authentication/EnterEmailForm.js
```diff
@@ -0,0 +1,217 @@
...
+
+  onMutationError(response) {
+    logger.error(response)
+    this.setState({
+      savingEmailInProgress: false,
+      otherError: true,
+    })
+  }
+
+  handleKeyPress(e) {
```
amaliwanag2, line 135: i'm pretty sure it's possible to trigger this function twice before your set state call finishes.  You can defend against this using lodash debounce on this function

web/src/js/components/Authentication/EnterEmailForm.js
```diff
@@ -0,0 +1,217 @@
...
+        }}
+      >
+        <EmailField
+          data-test-id={'enter-email-form-email-field'}
+          ref={elem => {
+            this.emailFieldRef = elem
+          }}
+          emailDuplicate={this.state.emailDuplicate}
+          otherError={this.state.otherError}
+          onKeyPress={this.handleKeyPress.bind(this)}
```
amaliwanag2, line 169: @jedtan Tried testing this locally and on dev (verified that this code is on dev) and when I try to submit my new email it logs me out without updating.  I think my config might be slightly off on local, can you give me steps to test?

jedtan: Hmm, I don't think I had to do anything, I still have my default test user is the "abcdefghijklmno" one - did it work for you when it was on dev?

web/src/js/components/Settings/Account/AccountComponent.js
```diff
@@ -87,6 +107,27 @@ class Account extends React.Component {
     this.setState({ usernameUpdated: true })
   }
 
+  openEmailDialog() {
+    console.log(this.state)
+    if (!this.state.emailReauthed) {
+      replaceUrl(loginURL, {
+        next: constructUrl(
+          '/newtab/apple',
+          { reauthed: true }
```
jedtan, line 116: This is not showing up in the URL and I can't figure out why

web/src/js/authentication/user.js
```diff
@@ -330,3 +335,24 @@ export const reloadUser = async () => {
...
+      .verifyBeforeUpdateEmail(newEmail, {
+        url: continueUrl,
+      })
+      .then(function() {
+        resolve()
+      })
+      .catch(function(error) {
+        reject(error)
+      })
+  })
```
amaliwanag2, line 357: stylistic info: in an async/await context this promise could be written as just
`const user = firebase.auth().currentUser`
`await user.verifyBeforeUpdateEmail(newEmail, {        url: continueUrl,      })`
if there is an error in 'verifyBeforeUpdateEmail', the updateUserEmail promise will be reject
actually since async wraps and returns a promise you could also write this block as 
`const user = firebase.auth().currentUser`
`return user.verifyBeforeUpdateEmail(newEmail, {        url: continueUrl,      })`

web/src/js/components/General/EmailField.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  setErrorMessage(message) {
+    this.setState({
+      validationErrorMessage: message,
+    })
+  }
+
+  validate() {
+    if (this.hasValue()) {
```
amaliwanag2, line 37: we may want to add a basic email validation library or regex on the front end
something that checks that the email has an @ and a .x

web/src/js/components/General/EmailField.js
```diff
@@ -0,0 +1,82 @@
...
+    super(props)
+    this.inputElem = null
+    this.state = {
+      email: null,
+      validationErrorMessage: null,
+    }
+  }
+
+  hasValue() {
+    return !!(
```
amaliwanag2, line 16: just curious what the type coercion is for?

web/src/js/components/Settings/Account/EnterEmailForm.js
```diff
@@ -0,0 +1,175 @@
...
+      this.setState({ emailDuplicate: true })
+    } else if (error.code === 'auth/internal-error') {
+      const errorMessage = JSON.parse(error.message)
+      if (errorMessage.error.message === 'INVALID_NEW_EMAIL')
+        this.setState({ emailInvalid: true })
+    }
+  }
+
+  submit() {
+    const email = this.emailFieldRef.getValue()
```
amaliwanag2, line 91: this paradigm works, there's another way to do this that also has some benefits 

On the email input you can add a debounced onChange function that sets the email input value to state
The benefits of reading input from state are:
- you can read from state to validate
- you can make the 'submit' button disabled until it's' a valid email
- you stay in the synthetic react dom which is encouraged
I think this article takes a stronger stance than I have personally but it talks about some of the drawbacks of refs https://blog.logrocket.com/why-you-should-use-refs-sparingly-in-production/

if you do switch over to updating and reading from state you'll definitely want to debounce so you don't trigger a rerender every single time the user changes the input value

kmjennison (changes requested):
Functionally, looks great. Just want to get some more test coverage.

web/src/js/components/General/EmailField.js
jedtan: To get it back to a boolean, but I do think it would technically work without this coercion.

web/src/js/components/Settings/Account/__tests__/AccountComponent.test.js
```diff
@@ -285,4 +293,66 @@ describe('Account component', () => {
...
+    const emailButton = buttons.at(1)
+    emailButton.simulate('click')
+
+    expect(replaceUrl).toHaveBeenCalledWith(loginURL, {
+      next: constructUrl(accountURL, { reauthed: true }),
+      reauth: 'true',
+    })
+  })
+
+  it('redirects on clicking the check email form if there is no reauthed parameter', () => {
```
kmjennison, line 314: Inaccurate test description?

web/src/js/components/Settings/Account/__tests__/AccountComponent.test.js
```diff
@@ -285,4 +293,66 @@ describe('Account component', () => {
...
+    const userData = getMockUserData()
+    delete userData.email
+
+    getUrlParameters.mockReturnValue({ verified: true })
+
+    const wrapper = mount(<AccountComponent user={userData} />)
+
+    let state = wrapper.state()
+    expect(state.emailOpen).toEqual(true)
+    expect(state.emailVerified).toEqual(true)
```
kmjennison, line 348: General feedback (no need to change here): asserting against component state and other internals is not ideal. Rather, focus tests on rendering/output like you do with the assertion below.

kmjennison (approved):

PR #909, opened 2021-04-19
add isv4 to revenue and tab log

amaliwanag2:


amaliwanag2:
@kmjennison verified on dev aws db that legacy logs isV4 as false and cats logs isV4 is true on both userTabsLog table and UserRevenueLog tabe

kmjennison (approved):
LGTM 🍝

PR #913, opened 2021-04-20
Set cookie if v4BetaEnabled is true

amaliwanag2:


kmjennison (approved):
LGTM 🍝  -- I'm deploying to dev to manually test, too.

PR #924, opened 2021-05-04
Alec/tfac 364 - make end to end tests only run on master

amaliwanag2:


.travis.yml
```diff
@@ -248,12 +248,12 @@ install: yarn run all:install
...
-script: >
-  yarn run ci:test &&
-  if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
-    yarn run ci:deploy test &&
-    yarn run ci:test:integration
-  fi
+script:
+  - 'yarn run ci:test'
+  - 'if [ ${TRAVIS_BRANCH} = "master" ]; then
+      yarn run ci:deploy test || travis_terminate 1;
```
amaliwanag2, line 254: The if block is executed entirely as one command, and the exit code is checked after it is finished.  travis_terminate 1 fails the build immediately if either of these yarn commands fail

amaliwanag2:
@kmjennison verified by making the conditional check run against dev then testing that a failed end to end test stops the deploy

kmjennison (approved):
LGTM! 🍝

PR #965, opened 2021-05-25
create invitedUsers table

amaliwanag2:
This is what I'm thinking for the invitedUsers table 
- I added a field for whether it's a squad invite because I think we may want to use it with a new Squad model but we don't have to include it for now

 **next steps are**
- create a inviteUser mutation that creates a new invitedUser, updates the user model with a 'name' field
- create a getInvitedUsers mutation that gets all invited users for a given user id

**steps after that**
- add the sendgrid api to the invite user mutation
- trigger the sending of a templated email

**final steps**
- make email give a link that takes users to extension?  Or just landing page with a new kind of referral type?  we'll need to think through this a little but I'll also need to make sure that the userReferral and invitedUsers types are both updated when a user signs up.

**Question**:  can a user be already email verified if they sign up through email invite?


graphql/database/invitedUsers/InvitedUsersModel.js
```diff
@@ -0,0 +1,75 @@
...
+        .description(
+          `whether the user is invited to a v4 squad or is a normal referral`
+        ),
+      status: types
+        .string()
+        .allow('pending', 'accepted')
+        .default(self.fieldDefaults.status) // only set in app code
+        .description(
+          `the status of the invite. A user is pending and then can accept`
+        ),
```
kmjennison, line 52: Add `invitedUserId` (set when the user joins), because their email may change over time

graphql/database/invitedUsers/InvitedUsersModel.js
```diff
@@ -0,0 +1,75 @@
...
+        .string()
+        .required()
+        .description(`the email address of a newly invited user`),
+      isSquadInvite: types
+        .boolean()
+        .default(self.fieldDefaults.isSquadInvite)
+        .description(
+          `whether the user is invited to a v4 squad or is a normal referral`
+        ),
+      status: types
```
kmjennison, line 46: Take or leave this feedback: if the status is only two states, it saves on storage to use an `accepted` boolean.

graphql/database/invitedUsers/InvitedUsersModel.js
kmjennison: Probably unnecessary until we have a squads, but we may eventually want some ID to identify the squad/invite. One benefit is that we'll still find the inviter link if the invited user chooses to sign up with a different email.

graphql/database/invitedUsers/InvitedUsersModel.js
```diff
@@ -0,0 +1,75 @@
...
+        .string()
+        .required()
+        .description(
+          `The unique user ID from our authentication service (Firebase)`
+        ),
+      invitedEmail: types
+        .string()
+        .required()
+        .description(`the email address of a newly invited user`),
+      isSquadInvite: types
```
kmjennison, line 40: Take or leave this feedback: I agree this is a bit future-building and out of this feature scope, but fine to leave too.

kmjennison (changes requested):
Recommended adding one field, plus some other comments with optional changes.

kmjennison:
@amaliwanag2 

> make email give a link that takes users to extension? Or just landing page with a new kind of referral type? we'll need to think through this a little 

Given we can use the `u` parameter for the inviter's username, I don't think we have to make any changes to the homepage flow, right? Let's keep it simple and send the user to the landing page for now.

> an a user be already email verified if they sign up through email invite?

This is potentially doable with the Firebase API but likely not worth the effort now and has some complications, such as a the "verify email" token expiring after ~30 mins.

graphql/database/users/logEmailVerified.js
```diff
@@ -29,6 +34,21 @@ const logEmailVerified = async (userContext, userId) => {
   if (userContext.emailVerified) {
     try {
       await rewardReferringUser(userContext, userId)
+      const override = getPermissionsOverride(ADMIN_MANAGEMENT)
+      const originalInvitedUsers = await InvitedUsersModel.query(
```
amaliwanag2, line 38: haven't tested this part deployed yet

graphql/database/users/logEmailVerified.js
```diff
@@ -29,6 +34,21 @@ const logEmailVerified = async (userContext, userId) => {
...
+        override,
+        userContext.email
+      )
+        .usingIndex('InvitesByInvitedEmail')
+        .execute()
+      await Promise.all(
+        originalInvitedUsers.map(invitedUserEntry =>
+          InvitedUsersModel.update(override, {
+            ...invitedUserEntry,
+            invitedId: userId,
```
amaliwanag2, line 48: if a person has been invited multiple times over a few months, do we invite all or just the latest?

graphql/database/invitedUsers/createInvitedUsers.js
```diff
@@ -0,0 +1,70 @@
...
+import xssFilters from 'xss-filters'
+import InvitedUsersModel from './InvitedUsersModel'
+import UserModel from '../users/UserModel'
+import { verifyAndSendInvite } from './utils'
+/**
+ * conditionally creates a new invited user and sends a sendgrid email if user has not already
+ * been invited OR if inviting user has exceeded the max amount of invites in 24 hours
+ * @param {object} userContext - The user authorizer object.
+ * @param {string} inviterId - The user id.
+ * @param {any} invitedEmails - A UUID for the charity
```
kmjennison, line 12: Outdated description

graphql/database/invitedUsers/createInvitedUsers.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {object} inviterName - options to update impact record, include logImpact, claimPendingUserReferralImpact, confirmImpact
```
kmjennison, line 13: Outdated description

graphql/database/invitedUsers/createInvitedUsers.js
```diff
@@ -0,0 +1,70 @@
...
+  invitedEmails,
+  inviterName
+) => {
+  const endTimeRoundedISO = moment()
+    .startOf('hour')
+    .toISOString()
+  const startTimeRoundedISO = moment(endTimeRoundedISO)
+    .subtract(30, 'days')
+    .endOf('hour')
+    .toISOString()
```
kmjennison, line 28: Not too consequential, but I don't see the value in rounding to nearest hour here.

graphql/database/invitedUsers/utils.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+  if (hasUserBeenInvitedRecently.length) {
+    return { email: inviteEmail, error: 'user has been invited recently' }
+  }
+  const msg = {
+    to: inviteEmail,
+    from: invitingUser.email,
+    templateId: 'd-e974f14ae104424cb8237b234d485804',
+    dynamicTemplateData: {
+      name: inviterName,
```
kmjennison, line 46: Sanitize `inviterName`.

graphql/database/users/logEmailVerified.js
kmjennison: I think most useful to just update the one that was successful.

graphql/database/users/logEmailVerified.js
```diff
@@ -29,6 +34,21 @@ const logEmailVerified = async (userContext, userId) => {
...
+          })
+        )
+      )
```
kmjennison, line 51: Add tests for this. (Our code coverage GitHub bot isn't working right now. Not sure why and need to fix.)

graphql/database/invitedUsers/createInvitedUsers.js
```diff
@@ -0,0 +1,70 @@
...
+  const [invitingUser, userInvites] = await Promise.all([
+    UserModel.get(userContext, inviterId),
+    InvitedUsersModel.query(userContext, inviterId)
+      .usingIndex('InvitesByInviter')
+      .where('created')
+      .between(startTimeRoundedISO, endTimeRoundedISO)
+      .execute(),
+  ])
+  if (userInvites.length + sanitizedEmails.length > 50) {
+    throw new Error('user is trying to invite too many people in 24 hours')
```
kmjennison, line 41: Later, if we end up taking different actions based on error types, switch to error codes.

kmjennison (changes requested):
A few minor changes. After those, feel free to merge. Great job!

graphql/database/users/logEmailVerified.js
amaliwanag2: manually tested that this works

graphql/database/invitedUsers/createInvitedUsers.js
```diff
@@ -0,0 +1,86 @@
...
+  inviterName,
+  inviterMessage
+) => {
+  try {
+    const override = getPermissionsOverride(ADMIN_MANAGEMENT)
+    const endTimeRoundedISO = moment()
+      .startOf('hour')
+      .toISOString()
+    const startTimeRoundedISO = moment(endTimeRoundedISO)
+      .subtract(30, 'days')
```
kmjennison, line 35: Should be 24 hours

graphql/database/invitedUsers/createInvitedUsers.js
```diff
@@ -0,0 +1,86 @@
...
+  userContext,
+  inviterId,
+  invitedEmails,
+  inviterName,
+  inviterMessage
+) => {
+  try {
+    const override = getPermissionsOverride(ADMIN_MANAGEMENT)
+    const endTimeRoundedISO = moment()
+      .startOf('hour')
```
kmjennison, line 32: No need to round the hour, right?

graphql/database/invitedUsers/createInvitedUsers.js
```diff
@@ -0,0 +1,86 @@
...
+      .toISOString()
+    const startTimeRoundedISO = moment(endTimeRoundedISO)
+      .subtract(30, 'days')
+      .endOf('hour')
```
kmjennison, line 36: No need to round the hour, right?

kmjennison (changes requested):
Just a few fixes, otherwise looks great!

graphql/database/invitedUsers/createInvitedUsers.js
amaliwanag2: no need, i just like comparing times with 0s.

kmjennison (approved):

graphql/database/invitedUsers/utils.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+  if (hasUserBeenInvitedRecently.length) {
+    return { email: inviteEmail, error: 'user has been invited recently' }
+  }
+  const msg = {
+    to: inviteEmail,
+    from: invitingUser.email,
+    templateId: 'd-69707bd6c49a444fa68a99505930f801',
+    dynamicTemplateData: {
+      name: inviterName,
```
kmjennison, line 48: We need to URI encode the username in the URL param, or it will fail for some users. See [invite friends](https://github.com/gladly-team/tab/blob/master/web/src/js/components/Settings/Profile/InviteFriendComponent.js#L29) logic for an example.

kmjennison (changes requested):

kmjennison (approved):

PR #984, opened 2021-06-29
[TFAC-423] Update search provider

kmjennison:
After deploying to prod, follow up with an update to the prod search Lambda@Edge function version.


amaliwanag2 (approved):
:cheese: looks good to me

PR #985, opened 2021-07-01
Alec/tfac 412

amaliwanag2:


web/src/js/components/Settings/Widgets/WidgetSettingsComponent.js
```diff
@@ -19,15 +19,22 @@ class WidgetSettings extends React.Component {
...
     if (widget && widget.config) {
       config = JSON.parse(widget.config)
     }
 
-    const settings = this.getConfig(JSON.parse(appWidget.settings), config)
-
+    let settings = this.getConfig(JSON.parse(appWidget.settings), config)
+    // temporarily adding yahoo as an option for yahoo demo team
+    if (user.id === YAHOO_USER_ID) {
+      settings = settings.map(setting => {
```
amaliwanag2, line 31: I didn't bother with figuring out how to modify prod db.  this will only run on yahoo user

web/src/js/components/Settings/Widgets/WidgetSettingsComponent.js
```diff
@@ -19,15 +19,22 @@ class WidgetSettings extends React.Component {
...
+        if (setting.field === 'engine') {
+          setting.choices.push('Yahoo')
```
amaliwanag2, line 33: you'll have to make sure you've clicked the "yahoo" option for the demo lol

web/src/js/constants.js
```diff
@@ -126,3 +126,9 @@ export const SEARCH_SRC_OWN_PAGE = 'self'
 
 export const SEARCH_INTRO_QUERY_ENGLISH =
   'How many searches do people make every day?'
+
+export const STORAGE_YAHOO_SEARCH_DEMO = 'tab.yahoo.searchdemo'
+// local testing
+export const YAHOO_USER_ID = 'VXNlcjpjTDVLY0ZLSGQ5ZkVVNUM5VnN0ajNnNEpBYzcz'
```
amaliwanag2, line 132: @kmjennison you'll want to test this locally by pulling down my branch then setting your user id here

amaliwanag2: We can also make this a build env variable as well so we can change it based on env.

web/src/js/components/Widget/Widgets/Search/Search.js
```diff
@@ -41,7 +62,19 @@ class Search extends React.Component {
     const engine = this.state.config.engine || 'Google'
     const searchApi = this.getSearchApi(engine)
     const searchTerm = this.searchInput.input.value
-
+    //refetch latest local storage
+    const yahooHasAcknowledged = localStorageMgr.getItem(
+      STORAGE_YAHOO_SEARCH_DEMO
+    )
+    if (yahooHasAcknowledged === 'true') {
+      LogTabMutation({
```
kmjennison, line 70: This should be `LogSearchMutation`.

kmjennison: Use "tab" for the "source" input for that mutation.

web/src/js/components/Settings/Widgets/WidgetSettingsComponent.js
kmjennison: This will probably confuse the testers, so I'll propose a different approach.

web/src/js/constants.js
```diff
@@ -126,3 +126,9 @@ export const SEARCH_SRC_OWN_PAGE = 'self'
...
+// prod
+// export const YAHOO_USER_ID='Eh1S0gtEvYSWZjZ8B0kk4jk6zBm2'
```
kmjennison, line 134: Update this to: `VXNlcjpFaDFTMGd0RXZZU1daalo4QjBrazRqazZ6Qm0y`

web/src/js/constants.js
kmjennison: Fine to leave as-is and test this live on prod.

web/src/js/components/Dashboard/DashboardComponent.js
```diff
@@ -631,6 +631,7 @@ class Dashboard extends React.Component {
           user={user}
           isCampaignLive={showCampaign}
           showError={this.showError.bind(this)}
+          tabId={tabId}
```
kmjennison, line 634: Won't need to pass this with the other mutation (see below).

web/src/js/components/Widget/Widgets/Search/Search.js
```diff
@@ -41,7 +62,19 @@ class Search extends React.Component {
     const engine = this.state.config.engine || 'Google'
```
kmjennison, line 62: Here, if isYahooUser, set it to always be Yahoo. Then, we don't have to worry about the widget settings at all.

web/src/js/components/Settings/Widgets/WidgetSettingsComponent.js
kmjennison: Can revert these changes and rely only on the search component itself.

kmjennison (changes requested):

amaliwanag2:
@kmjennison tested again locally that changes have no impact on non yahoo user

web/src/js/components/Widget/Widgets/Search/Search.js
```diff
@@ -38,12 +58,27 @@ class Search extends React.Component {
...
-
-    // The page might be iframed, so opening in _top is critical.
-    window.open(searchApi + searchTerm, '_top')
+    //refetch latest local storage
+    const yahooHasAcknowledged = localStorageMgr.getItem(
+      STORAGE_YAHOO_SEARCH_DEMO
+    )
+    if (yahooHasAcknowledged === 'true') {
+      LogSearchMutation({
+        userId: this.props.user.id,
```
kmjennison, line 72: This isn't working because the mutation, counterintuitively, takes the raw user ID rather than the Relay global. That's bad mutation design on my part as a workaround for another search landing page use case.

See [here](https://github.com/gladly-team/tab/blob/master/web/src/js/components/Search/SearchResults.js#L202) for an example of how to get it working. In our case, I think you'll want to fetch the user ID prior to the search being executed (probably on component mount) because the promise resolution might be canceled on page navigation (unsure).

kmjennison (changes requested):

web/src/js/components/Widget/Widgets/Search/Search.js
amaliwanag2: @kmjennison  locally verified that the mutation logs to table "userSearchLog".  I think that's right?  I am not currently awaiting the return of this mutation because I figure we want it to redirect asap, and regardless of whether the search log mutation returns back 200.  Is that ok?

kmjennison (approved):
LGTM 🦊

PR #987, opened 2021-07-06
Add GAM "dev" key-value to allow testing custom ad campaigns

kmjennison:
The goal of this PR is to conditionally pass a "dev" key-value to Google Ad Manager when in our dev environment, which supports TFAC-393.


amaliwanag2 (approved):
looks good to me - gl hf with GAM

PR #999, opened 2021-07-20
TFAC-445 add cat day env notification

amaliwanag2:
didnt know if i should use a time library to calculate if it was august 8th or if we'd just push the change manually


kmjennison (changes requested):
Per in-person discussion.

kmjennison (approved):
LGTM 🦊

PR #1003, opened 2021-07-20
TFAC-434 Create Missions Mutation and TFAC-442 Create Squad Invite Mutation

amaliwanag2:


graphql/data/schema.js
```diff
@@ -1412,6 +1413,25 @@ const createInvitedUsersMutation = mutationWithClientMutationId({
...
+ */
+const createNewMissionMutation = mutationWithClientMutationId({
+  name: 'CreateNewMission',
+  inputFields: {
+    userId: { type: new GraphQLNonNull(GraphQLString) },
+    squadName: { type: new GraphQLNonNull(GraphQLString) },
+  },
+  outputFields: {
+    squadId: {
+      type: GraphQLString,
```
kmjennison, line 1427: Make non-null

graphql/database/missions/createMission.js
```diff
@@ -0,0 +1,45 @@
...
+  // user should not have been able to create a new mission
+  if (user.currentMissionId) {
+    throw new Error(
+      'attempting to create a new squad while still in an ongoing missions'
+    )
+  }
+  const missionId = nanoid(9)
+  const newMission = {
+    id: missionId,
+    squadName,
```
kmjennison, line 29: Let's limit the length of this. Limit characters with validator.js (semi-sanitize)? Not sure, LMK what you think.

kmjennison (changes requested):

graphql/database/missions/createMission.js
amaliwanag2: sounds good.  It seems more important to do this on the front end so they can't even create a squad with character length over 30, and if they hack it and somehow make a longer squad name there aren't really any repercussions to our data storage, but might as well do it on the backend as well.  Will also do the xss filter

graphql/data/schema.js
```diff
@@ -1415,6 +1416,66 @@ const createInvitedUsersMutation = mutationWithClientMutationId({
...
+          name: 'successfulSquadEmailAddresses',
+          fields: () => ({ email: { type: GraphQLString } }),
+        })
+      ),
+    },
+    existingUserInvited: {
+      type: new GraphQLList(
+        new GraphQLObjectType({
+          name: 'successfulExistingUsers',
+          fields: () => ({ username: { type: GraphQLString } }),
```
kmjennison, line 1443: My concern here is that a client can use this to arbitrarily map email addresses to usernames. It's not a security risk per se, but it's unnecessary data exposure that could cause problems in the future. Will we use these output fields on the client? If so, can we simply return successful/unsuccessful emails (only giving back data that the client gave us)?

kmjennison (changes requested):

graphql/data/schema.js
amaliwanag2: this makes sense.  We do end up using the  successful existing users usernames in the pending mission table, although we probably get them by refetching the current mission so it's not needed in this response.  I'm currently still limiting the amount of email invites someone can do in 24 hours (can reduce it even more) which should limit how many people someone can ping for.

kmjennison: My suggestion is that we only display the (pending) email address until a user joins the mission, at which point we show their username.

amaliwanag2: ok that works!

kmjennison (approved):
LGTM 🦊

PR #1006, opened 2021-07-22
TFAC-439 UpdateMissionNotification Mutation

jedtan:


graphql/database/missions/UserMissionModel.js
```diff
@@ -13,17 +13,17 @@ class UserMission extends BaseModel {
   }
 
   static get hashKey() {
-    return 'missionId'
+    return 'userId'
```
amaliwanag2, line 16: this makes sense, but we more often query and group UserMissions by missionId through the flow of the feature.  Also we would have to delete the tables in prod and recreate them.   I'd say let's keep the indexes as they are for now.

graphql/database/missions/updateMissionNotification.js
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Records a user's closing the "Mission Completed" notification.
+ *
+ * @param {object} userContext - The user authorizer object.
+ * @param {string} userId - The user's ID.
+ * @param {string} missionId - The mission's ID.
+ */
+const updateMissionNotification = async (userContext, userId, missionId) => {
+  // Get the mission model and modify it.
+  await UserMissionModel.update(userContext, {
```
amaliwanag2, line 12: instead of using UserContext here, you'll want to use
```
import {
  getPermissionsOverride,
  MISSIONS_OVERRIDE,
} from '../../utils/permissions-overrides'

const override = getPermissionsOverride(MISSIONS_OVERRIDE)
```
and pass in the override as the first argument to UserMissionModel.update

web/src/js/mutations/UpdateMissionNotificationMutation.js
```diff
@@ -0,0 +1,38 @@
+import graphql from 'babel-plugin-relay/macro'
```
amaliwanag2, line 1: we won't have this feature in legacy so we can remove this file from the web mutations for now.  We'll need it in tab-web so worth copying elsewhere and holding onto it.

graphql/database/missions/MissionModel.js
```diff
@@ -99,6 +100,16 @@ class Mission extends BaseModel {
       endOfMissionAwards: [],
     }
   }
+
+  static get permissions() {
```
amaliwanag2, line 104: this permissions block isn't needed with the override

graphql/database/missions/updateMissionNotification.js
```diff
@@ -0,0 +1,23 @@
+import UserMissionModel from './UserMissionModel'
+
+/**
+ * Records a user's closing the "Mission Completed" notification.
+ *
+ * @param {object} userContext - The user authorizer object.
+ * @param {string} userId - The user's ID.
+ * @param {string} missionId - The mission's ID.
+ */
+const updateMissionNotification = async (userContext, userId, missionId) => {
```
amaliwanag2, line 10: sorry I was vague in the ticket.  Let's modify this mutation to either update `acknowledgedMissionComplete` or `acknowledgedMissionStarted` fields on the userMission

graphql/data/schema.js
```diff
@@ -2066,6 +2067,22 @@ const deleteUserMutation = mutationWithClientMutationId({
   },
 })
 
+const updateMissionNotificationMutation = mutationWithClientMutationId({
+  name: 'UpdateMissionNotification',
+  inputFields: {
+    userId: { type: new GraphQLNonNull(GraphQLString) },
+    missionId: { type: new GraphQLNonNull(GraphQLString) },
+  },
```
amaliwanag2, line 2075: could add something like updatefield: { type graphqlEnum} with acknowledgeMissionStarted and acknowledgeMissionCompleted

amaliwanag2 (changes requested):

graphql/database/missions/UserMissionModel.js
jedtan: The reason I did this is for permission to update the UserMissionModel (e.g. the `userIdMatchesHashKey` permission authorizer). It sounds like we may need to write a new authorizer?

graphql/database/missions/updateMissionNotification.js
jedtan: Ohh

web/src/js/mutations/UpdateMissionNotificationMutation.js
jedtan: Ok!

graphql/database/missions/MissionModel.js
jedtan: Cool, sounds good.

graphql/data/schema.js
jedtan: sounds good.

graphql/database/missions/updateMissionNotification.js
amaliwanag2: @jedtan ok wow I see my two tickets were kind of open ended, and I see you have a WIP for when a user accepts a mission with the additional logic for moving a user to accepted in the missionDocument.  Up to you if you want to combine the PRs and mutations or keep them seperate.

jedtan: Eh I think separate is okay. They do pretty different things imo

amaliwanag2 (approved):
👍 LGTM.  I'm guessing the next pr will add to the same mutation

PR #1011, opened 2021-07-30
[wip] Jtan/tfac 437

jedtan:


graphql/database/missions/UserMissionModel.js
```diff
@@ -64,10 +64,27 @@ class UserMission extends BaseModel {
...
         .description(`count of the current tab streak in current mission`),
       missionMaxTabsDay: types
-        .number()
-        .integer()
-        .default(self.fieldDefaults.missionMaxTabsDay)
-        .description(`most tabs in a single day during mission`),
+        .object({
+          // The count of tabs for the day on which the user opened
+          // the most tabs.
+          maxDay: types.object({
```
amaliwanag2, line 70: i like the nested attribute.  Can you do this with tab streak as well?

graphql/database/users/logTab.js
```diff
@@ -75,28 +75,28 @@ const logTab = async (userContext, userId, tabId = null, isV4 = true) => {
...
+
   const todayTabCount = getTodayTabCount(user) + 1
   const isValid = isTabValid(todayTabCount, user.lastTabTimestamp)
 
-  // Update the user's counter for max tabs in a day.
-  // If this is the user's first tab today, reset the counter
-  // for the user's "current day" tab count.
-  // If today is also the day of all time max tabs,
-  // update the max tabs day value.
-  const isTodayMax = todayTabCount >= user.maxTabsDay.maxDay.numTabs
```
amaliwanag2, line 86: @jedtan we'll still want to calculate and store a user's alltime max tab count.  This is to support legacy users

amaliwanag2: sorry disregard, I see you do it below

graphql/database/users/logTab.js
```diff
@@ -75,28 +75,28 @@ const logTab = async (userContext, userId, tabId = null, isV4 = true) => {
...
-      numTabs: isTodayMax ? todayTabCount : user.maxTabsDay.maxDay.numTabs,
-    },
-    recentDay: {
-      date: moment.utc().toISOString(),
-      numTabs: todayTabCount,
-    },
+  if (user.has('currentMissionId') && isValid) {
+    mission = await MissionModel.get(userContext, currentMissionId)
+    if (!mission.has('completed')) {
+      userMission = await UserMissionModel.query(userContext, currentMissionId, userId)
```
amaliwanag2, line 85: @jedtan lmk if you have questions setting up dummy data in your local dynamodb

amaliwanag2: also you'll need to user an override on these.  I think you can just user UserMissionModel.get(override,missionId,userId) here.  If using query, you'll need to call usingIndex(index)..execute() in order for the query to return anything

jedtan:
just a draft right now!

amaliwanag2:
already merged

PR #1019, opened 2021-08-09
[TFAC-437] Adding code to Complete a Mission

jedtan:


graphql/database/missions/__tests__/utils.test.js
```diff
@@ -143,6 +143,7 @@ const mockDefaultMissionReturn = {
   tabCount: 258,
   squadMembers: [
     {
+      userId: 'cL5KcFKHd9fEU5C9Vstj3g4JAc73',
```
amaliwanag2, line 146: this is mocking the return value we return to the client, where we hide userIds.  It looks like you're no longer using this value either

graphql/database/missions/completeMission.js
```diff
@@ -0,0 +1,63 @@
...
+  const mostDayTabsMember = mission.squadMembers.reduce((prev, current) =>
+    prev.missionMaxTabsDay > current.missionMaxTabsDay ? prev : current
+  )
+  const mostTotalTabsMember = mission.squadMembers.reduce((prev, current) =>
+    prev.tabs > current.tabs ? prev : current
+  )
+
+  const endOfMissionAwards = [
+    {
+      awardType: 'Consistent Kitty',
```
amaliwanag2, line 40: lets make these constants

amaliwanag2 (changes requested):
Just one small change.  Otherwise, Looks good to me!

graphql/database/missions/__tests__/utils.test.js
jedtan: I'm not sure I follow - I did amend `getCurrentUserMission` but it sounds like I shouldn't do that? I was adding it so I would have the userIds of squad members when calculating end of mission awards.

amaliwanag2: actually that's' fine, graphql will still mask the user ids on calls to the FE so disregard

amaliwanag2 (approved):
LGTM

PR #1039, opened 2021-08-28
hasSeenSquads Backend Work

jedtan:


amaliwanag2:
@jedtan oops I already did this work, and added some additional logic that sets has seen squads to true if a user was invited to a sqaud.

PR #1052, opened 2021-09-01
slightly adjust log tab and add more tests

amaliwanag2:
log tab was counting tabs while mission was in pending state


kmjennison (approved):
LGTM 🦊

PR #1053, opened 2021-09-01
start mission once second user joins

amaliwanag2:
we weren't actually kicking off the mission when a second user joins lol


kmjennison (approved):
LGTM 🦊

PR #1072, opened 2021-09-22
add truex id to user model

amaliwanag2:
verified on local front end that truex id is passed down and stays consistent between calls


graphql/data/schema.js
```diff
@@ -459,7 +460,7 @@ const userType = new GraphQLObjectType({
     backgroundImage: {
       type: backgroundImageType,
       description: "Users's background image",
-      resolve: (user, args, context) => getBackgroundImage(context.user, user),
+      resolve: (user, _args, context) => getBackgroundImage(context.user, user),
```
amaliwanag2, line 463: fyi if you didn't know that latest eslint rules want _ prefix if variable isn't used.  I know this is old code but just learned that recently -- similar to python

kmjennison (changes requested):
Test are broken. Please ping me to review when fixed.

kmjennison (approved):
LGTM 🦊

PR #1080, opened 2021-10-13
Alec/tfac 579 580 595

amaliwanag2:
**This should**

- prevent users from inviting existing users multiple times
- prevent the app from crashing if logging tabs in a squad gets corrupted
- fix the issue with kevin's user


kmjennison (approved):
LGTM! 🦊

PR #1086, opened 2021-10-16
TFAC-586 Adjusting schema.js for Tab for X Generalization

jedtan:

PR #1090, opened 2021-10-17
update schema to allow for querying for cause

amaliwanag2:
and relevant updates to user impact

You can now query for cause fields when developing locally and this code is on dev to verify that these changes do not break anything for current users

### DIFF
--- a/dynamodb/fixtures/Causes.json
+++ b/dynamodb/fixtures/Causes.json
@@ -1,6 +1,6 @@
 [
   {
-    "id": "123456789",
+    "id": "7f8476b9-f83f-47ac-8173-4a1c2ec3dc29",
     "charityId": "abcdefghijklmnop",
     "landingPagePath": "/test",
     "impactVisits": 10,
@@ -16,7 +16,7 @@
     },
     "theme": {
       "primaryColor": "#5094FB",
-      "secondayColor": "#29BEBA"
+      "secondaryColor": "#29BEBA"
     },
     "squads": {
       "squadCounterText": "squadCounterText",
@@ -30,8 +30,25 @@
       "missionCompleteSubtitle": "missionCompleteSubtitle",
       "impactCounterText": "impactCounterText"
     },
+    "sharing": {
+      "title": "sharingTitle",
+      "subtitle": "sharingSubtitle",
+      "imgCategory": "currentMissionDetails",
+      "redditButtonTitle": "redditButtonTitle",
+      "facebookButtonTitle": "facebookButtonTitle",
+      "currentMissionStep3": "currentMissionStep3",
+      "twitterButtonTitle": "twitterButtonTitle",
+      "tumblrTitle": "tumblrTitle",
+      "tumblrCaption": "tumblrCaption"
+    },
     "onboarding": {
-      "steps": [],
+      "steps": [
+        {
+          "title": "onboarding-title-1",
+          "subtitle": "onboarding-subtitle-1",
+          "imgName": "onboarding-imgName-1"
+        }
+      ],
       "firstTabIntroDescription": "firstTabIntroDescription"
     }
   }
--- a/graphql/data/schema.js
+++ b/graphql/data/schema.js
@@ -25,6 +25,7 @@ import {
   WIDGET,
   CHARITY,
   USER,
+  CAUSE,
   BACKGROUND_IMAGE,
   USER_IMPACT,
   USER_RECRUITS,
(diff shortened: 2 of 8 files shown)


dynamodb/fixtures/Causes.json
```diff
@@ -30,8 +30,25 @@
       "missionCompleteSubtitle": "missionCompleteSubtitle",
       "impactCounterText": "impactCounterText"
     },
+    "sharing": {
+      "title": "sharingTitle",
+      "subtitle": "sharingSubtitle",
+      "imgCategory": "currentMissionDetails",
+      "redditButtonTitle": "redditButtonTitle",
+      "facebookButtonTitle": "facebookButtonTitle",
+      "currentMissionStep3": "currentMissionStep3",
```
kmjennison, line 39: currentMissionStep3 shouldn't be here

graphql/data/schema.js
```diff
@@ -487,6 +491,10 @@ const userType = new GraphQLObjectType({
       description: 'a unique user ID sent to video ad partner truex',
       resolve: (user, _args, context) => getOrCreateTruexId(context.user, user),
     },
+    causeId: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: "the id of the user's current causue",
+    },
```
kmjennison, line 497: Don't need this. We already resolve in the `cause` property.

graphql/data/schema.js
```diff
@@ -697,6 +705,182 @@ const userType = new GraphQLObjectType({
   interfaces: [nodeInterface],
 })
 
+const causeType = new GraphQLObjectType({
+  name: CAUSE,
+  description: 'A person who uses our app',
```
kmjennison, line 710: Update description.

kmjennison (changes requested):
Bailed on the review early. Need to reconcile with:
https://github.com/gladly-team/tab/pull/1086/files

graphql/data/schema.js
```diff
@@ -166,6 +170,9 @@ const { nodeInterface, nodeField } = nodeDefinitions(
     if (type === VIDEO_AD_LOG) {
       return VideoAdLogModel.get(context.user, id)
     }
+    if (type === CAUSE) {
+      return getCause(context.user, context.user.id)
```
kmjennison, line 174: We will still want a node to resolve based on its global ID here. The question is more: do we need a new node definition for the object? If not—which is the case for Causes at the moment—we don't need this at all. But if we did want a node definition, we'd want stable IDs that resolve to every instance.

graphql/data/schema.js
```diff
@@ -2392,14 +2579,13 @@ const queryType = new GraphQLObjectType({
...
       },
       resolve: async (_, args, context) => {
-        const { userId, charityId } = args
-        return (await UserImpactModel.getOrCreate(context.user, {
-          userId,
-          charityId,
-        })).item
+        const { userId } = args
+        const userImpact = await getUserImpact(context.user, userId)
+        return userImpact
```
kmjennison, line 2588: Let's leave this as-is. Instead, add an impact resolver to the user, like `currentImpact`. You're right that this shouldn't be a root field. Now that it is, though, I think it should be able to resolve to any item in UserImpact.

graphql/data/schema.js
```diff
@@ -171,7 +171,7 @@ const { nodeInterface, nodeField } = nodeDefinitions(
       return VideoAdLogModel.get(context.user, id)
     }
     if (type === CAUSE) {
-      return getCause(context.user, context.user.id)
+      return CauseModel(context.user, id)
```
kmjennison, line 174: Need `CauseModel.get`

graphql/data/schema.js
```diff
@@ -484,6 +484,12 @@ const userType = new GraphQLObjectType({
       description: "Users's background image",
       resolve: (user, _args, context) => getBackgroundImage(context.user, user),
     },
+    userImpact: {
+      type: userImpactType,
+      description: "A user's charity specific impact",
```
kmjennison, line 489: Add "for the cause they're currently supporting"

kmjennison (changes requested):

graphql/database/cause/__tests__/fileContentsReader.test.js
```diff
@@ -0,0 +1,46 @@
...
+
+// jest.mock('path')
+jest.mock('fs')
+
+beforeEach(() => {
+  // join.mockImplementation((_, ...otherArgs) => {
+  //   const mockDirName = '/some/mock/directory/'
+  //   return [mockDirName, ...otherArgs].join('')
+  // })
+  // global.__dirname = 'foo'
```
kmjennison, line 14: Remove

PR #1098, opened 2021-10-20
set background image based on category in cause

amaliwanag2:


graphql/database/users/createUser.js
```diff
@@ -72,9 +68,10 @@ const createUser = async (
 
   // Set default background image to a cat image if user is enabled for v4 beta
   if (v4BetaEnabled) {
+    const { backgroundImageCategory } = await getCause(causeId)
```
amaliwanag2, line 71: merged in Homepage code that sends up the causeId.  I'm not fully sure about why this is idempotent but I think that createUser is always called from the auth front end and so that value should always be set in local storage.  Is this right?

amaliwanag2:
verified in dev

graphql/data/schema.js
```diff
@@ -2025,6 +2025,7 @@ const setUserBkgDailyImageMutation = mutationWithClientMutationId({
   name: 'SetUserBkgDailyImage',
   inputFields: {
     userId: { type: new GraphQLNonNull(GraphQLString) },
+    //  TODO - remove once we update FE to not send category ID up anymore
```
kmjennison, line 2028: In fact, we should leave this and assume some clients will be outdated (they will be).

graphql/database/users/createUser.js
kmjennison: No, it won't necessarily exist. `createUser` is called every time a user logs in, I believe, and their local storage could have cleared since they joined.

I'm wondering, can we just get rid of this logic here altogether and rely on the `getBackgroundImage` resolver to set an image (by category) if one is not already set? We already do a lot of this work elsewhere—no real reason for the data to be set on user creation. Thoughts? I might be missing something, so push back if you feel otherwise.

graphql/data/schema.js
kmjennison: This is where we'd want to formally deprecate the argument and then remove it in a number of months. (We haven't set a process for this yet.)

kmjennison (changes requested):

graphql/database/users/createUser.js
amaliwanag2: I liked this idea.  I pushed that move, I think it's pretty comprehensive.  Lmk what you think

graphql/database/cause/causes/cats/causeData.js
```diff
@@ -18,6 +18,7 @@ const data = {
   id: 'CA6A5C2uj',
   charityId: '6ce5ad8e-7dd4-4de5-ba4f-13868e7d212z', // Greater Good
   impactVisits: 14,
+  backgroundImageCategory: 'cats',
```
amaliwanag2, line 21: do relative imports work in this file.  Like can i import the constant value?

PR #1120, opened 2021-11-08
test new version of e2e tests and remove background image on auth

amaliwanag2:


amaliwanag2:
@kmjennison i tried upgrading jest to jest 27 for the new end to end tests release, but that ended up requiring migrating to CRA 4+ and I didn't want to deal with running migrations for this.  Looking at how Tab Homepage is currently running, we likely don't need to be on jest 27 to use the latest version of tab e2e but we won't have retries. 

homepage and tabweb will also need to update to tab-e2e 2.0.5 as I remove checking for the background image in the latest.

kmjennison (approved):
LGTM 🦊

PR #1138, opened 2021-11-18
update image file paths

amaliwanag2:


graphql/data/schema.js
```diff
@@ -1679,7 +1687,7 @@ const updateImpactMutation = mutationWithClientMutationId({
   name: 'UpdateImpact',
   inputFields: {
     userId: { type: new GraphQLNonNull(GraphQLString) },
-    charityId: { type: new GraphQLNonNull(GraphQLString) },
+    charityId: { type: GraphQLString },
```
amaliwanag2, line 1690: beginning deprecating sending up charity ID since we now get it from a user's cause

kmjennison (commented):
LGTM 🦊

PR #1152, opened 2022-01-16
TFAC-794 UserExperiment Model and Table

jedtan:


dynamodb/serverless.yml
```diff
@@ -434,6 +443,15 @@ custom:
           minimum: 1
           maximum: 40000
           usage: 0.60
+      userExperimentTable:
+        read:
+          minimum: 80
```
kmjennison, line 448: Set to 1. No plans to read from this table.

dynamodb/serverless.yml
```diff
@@ -434,6 +443,15 @@ custom:
...
+          maximum: 40000
+          usage: 0.75
+        write:
+          minimum: 20
```
kmjennison, line 452: Set to 1, which should be OK throughput to start, or it'll scale up as needed.

kmjennison (changes requested):
Just reduction to minimum table throughput on prod, then LGTM!

dynamodb/serverless.yml
jedtan: I think the small thing I wanted to comment on was that based on how we might need to design our ExperimentLog, we need to do an upsert of ExperimentLog since `trackingCallback` is called every time a user is placed into an experiment. "Placed" in Growthbook's terminology means every time they are exposed to the experiment (e.g. every time they see the landing page).

https://docs.google.com/document/d/1ru-oO7-OWVM3ByYZseJBQ-Mu8_1LwcLUMNlHdpQ2JN4/edit#

I can still scale this value down if we think it's high, but I think there's an argument to having it potentially be higher than 1 if, we are, say running tests on the Index page.

kmjennison: Ah, I see about the reads. That makes sense.

I do think these are too high, though—let's set the read/write minimum to 5. We want to give it room to drop down fairly low when usage drops.

kmjennison (changes requested):
See "offline" chat.

dynamodb/serverless.yml
jedtan: Actually, I think there's a way to avoid the reads, so we can do read ~5, write ~1.

jedtan: Just kidding, let's do 5/5. We can always change it in my later PR where we discuss the actual mutation implementation.

kmjennison (approved):
LGTM 🦊

PR #1153, opened 2022-01-21
Implementing UserSwitchSearchPromptLog Table

jedtan:


dynamodb/serverless.yml
```diff
@@ -396,6 +405,15 @@ custom:
           minimum: 40
           maximum: 40000
           usage: 0.60
+      userSwitchSearchPromptLogTable:
+        read:
+          minimum: 80
```
kmjennison, line 410: Set minimum to 1. Will be a very low-read table.

dynamodb/serverless.yml
```diff
@@ -396,6 +405,15 @@ custom:
...
+          maximum: 40000
+          usage: 0.75
+        write:
+          minimum: 40
```
kmjennison, line 414: Set minimum to 1. Will likely be a very low-write table.

kmjennison: FYI, the main reasons we have minimums is: 1) to better manage a spike in throughput when we enable a feature, because the DB will need a few minutes to scale up -- so we'll set it higher for a new table with expected, immediate high traffic; and 2) to be more resilient to spiky traffic for critical tables.

graphql/database/users/UserSwitchSearchPromptLogModel.js
```diff
@@ -0,0 +1,72 @@
+import moment from 'moment'
```
kmjennison, line 1: You'll also need to add IAM role permissions to `graphql/serverless.yml`.

graphql/database/users/UserSwitchSearchPromptLogModel.js
```diff
@@ -0,0 +1,72 @@
...
+      timestamp: moment.utc().toISOString(),
+    }
+  }
+
+  static get permissions() {
+    return {
+      get: permissionAuthorizers.userIdMatchesHashKey,
+      update: permissionAuthorizers.userIdMatchesHashKey,
+      create: permissionAuthorizers.userIdMatchesHashKey,
+      query: permissionAuthorizers.userIdMatchesHashKey,
```
kmjennison, line 65: I think our graphhql service will only need to create items. Other permissions unneeded.

kmjennison (changes requested):

dynamodb/serverless.yml
jedtan: Thank you for the explanation!

kmjennison (approved):
LGTM 🦊

PR #1167, opened 2022-02-04
Bumping some Dependencies

jedtan:


kmjennison:
@jedtan Does this fix local dev for you?

There's a compatibility issue. The options:
* Upgrade react-scripts to latest in the web service
  * Let's only deal with this if it's easy. You might give it a try for 20 mins, then if things look bad, drop it?
  * [Breaking in v4](https://github.com/facebook/create-react-app/blob/main/CHANGELOG-4.x.md#migrating-from-34x-to-400) / [breaking in v5](https://github.com/facebook/create-react-app/blob/main/CHANGELOG.md#breaking-changes)
* Just use `SKIP_PREFLIGHT_CHECK=true` -- fine by me if tests pass and things are stable

From the [build logs](https://app.travis-ci.com/github/gladly-team/tab/builds/245907178#L1091):

> There might be a problem with the project dependency tree.
> 
> It is likely not a bug in Create React App, but something you need to fix locally.
> 
> The react-scripts package provided by Create React App requires a dependency:
> 
>   "babel-jest": "^24.8.0"
> 
> Don't try to install it manually: your package manager does it automatically.
> 
> However, a different version of babel-jest was detected higher up in the tree:
> 
>   /home/travis/build/gladly-team/tab/node_modules/babel-jest (version: 27.4.0) 
> 
> Manually installing incompatible versions is known to cause hard-to-debug issues.
> 
> If you would prefer to ignore this check, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
> 
> That will permanently disable this message but you might encounter other issues.
> 
> To fix the dependency tree, try following the steps below in the exact order:
> 
>   1. Delete package-lock.json (not package.json!) and/or yarn.lock in your project folder.
> 
>   2. Delete node_modules in your project folder.
> 
>   3. Remove "babel-jest" from dependencies and/or devDependencies in the package.json file in your project folder.
> 
>   4. Run npm install or yarn, depending on the package manager you use.
> 
> In most cases, this should be enough to fix the problem.
> 
> If this has not helped, there are a few other things you can try:
> 
>   5. If you used npm, install yarn (http://yarnpkg.com/) and repeat the above steps with it instead.
> 
>      This may help because npm has known issues with package hoisting which may get resolved in future versions.
> 
>   6. Check if /home/travis/build/gladly-team/tab/node_modules/babel-jest is outside your project directory.
> 
>      For example, you might have accidentally installed something in your home folder.
> 
>   7. Try running npm ls babel-jest in your project folder.
> 
>      This will tell you which other package (apart from the expected react-scripts) installed babel-jest.
> 
> If nothing else helps, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
> 
> That would permanently disable this preflight check in case you want to proceed anyway.

kmjennison (changes requested):
See above comment.

PR #1180, opened 2022-02-19
TFAC-781 Resolvers for Getting User Search Engine

jedtan:
Also included, some refactoring, implementation of `getUserFeature` and some breaking up of shared code.

### DIFF
--- a/graphql/data/schema.js
+++ b/graphql/data/schema.js
@@ -121,7 +121,6 @@ import UserExperimentModel from '../database/experiments/UserExperimentModel'
 import getSearchEngine from '../database/search/getSearchEngine'
 import SearchEngineModel from '../database/search/SearchEngineModel'
 import getUserFeatures from '../database/experiments/getUserFeatures'
-import getUserFeature from '../database/experiments/getUserFeature'
 
 // eslint-disable-next-line import/no-named-as-default
 import getRecruits, {
@@ -137,7 +136,7 @@ import {
 } from '../database/globals/globals'
 import getCampaign from '../database/globals/getCampaign'
 import getUserSearchEngine from '../database/users/getUserSearchEngine'
-import { YAHOO_SEARCH_EXISTING_USERS } from '../database/experiments/experimentConstants'
+import getShouldShowYahooPrompt from '../database/users/getShouldShowYahooPrompt'
 
 class App {
   constructor(id) {
@@ -757,9 +756,7 @@ const userType = new GraphQLObjectType({
     showYahooPrompt: {
       type: new GraphQLNonNull(GraphQLBoolean),
       description: 'whether to show the yahoo search prompt',
-      resolve: user =>
-        !user.yahooSearchSwitchPrompt.hasSeenPrompt &&
-        getUserFeature(YAHOO_SEARCH_EXISTING_USERS).variation,
+      resolve: (user, _, context) => getShouldShowYahooPrompt(context, user),
     },
   }),
   interfaces: [nodeInterface],
--- a/graphql/data/schema.js
+++ b/graphql/data/schema.js
@@ -121,6 +121,7 @@ import UserExperimentModel from '../database/experiments/UserExperimentModel'
 import getSearchEngine from '../database/search/getSearchEngine'
 import SearchEngineModel from '../database/search/SearchEngineModel'
 import getUserFeatures from '../database/experiments/getUserFeatures'
+import getUserFeature from '../database/experiments/getUserFeature'
 
 // eslint-disable-next-line import/no-named-as-default
 import getRecruits, {
@@ -136,6 +137,7 @@ import {
 } from '../database/globals/globals'
 import getCampaign from '../database/globals/getCampaign'
 import getUserSearchEngine from '../database/users/getUserSearchEngine'
+import { YAHOO_SEARCH_EXISTING_USERS } from '../database/experiments/experimentConstants'
 
 class App {
   constructor(id) {
@@ -748,14 +750,16 @@ const userType = new GraphQLObjectType({
       resolve: (user, args, context) => getUserFeatures(context, user),
     },
     searchEngine: {
-      type: new GraphQLNonNull(GraphQLString),
+      type: SearchEngineType,
       description: 'the User’s search engine',
       resolve: user => getUserSearchEngine(user),
     },
     showYahooPrompt: {
       type: new GraphQLNonNull(GraphQLBoolean),
       description: 'whether to show the yahoo search prompt',
-      resolve: user => !user.yahooSearchSwitchPrompt.hasSeenPrompt,
+      resolve: user =>
+        !user.yahooSearchSwitchPrompt.hasSeenPrompt &&
+        getUserFeature(YAHOO_SEARCH_EXISTING_USERS).variation,
     },
   }),
   interfaces: [nodeInterface],
--- a/graphql/data/schema.js
+++ b/graphql/data/schema.js
@@ -752,7 +752,7 @@ const userType = new GraphQLObjectType({
     searchEngine: {
       type: SearchEngineType,
       description: 'the User’s search engine',
-      resolve: user => getUserSearchEngine(user),
+      resolve: (user, args, context) => getUserSearchEngine(context, user),
     },
     showYahooPrompt: {
       type: new GraphQLNonNull(GraphQLBoolean),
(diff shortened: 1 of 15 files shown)


graphql/data/schema.js
```diff
@@ -746,6 +747,16 @@ const userType = new GraphQLObjectType({
       description: 'feature values for this specific user',
       resolve: (user, args, context) => getUserFeatures(context, user),
     },
+    searchEngine: {
+      type: new GraphQLNonNull(GraphQLString),
```
kmjennison, line 751: This should be a SearchEngineType.

graphql/data/schema.js
```diff
@@ -746,6 +747,16 @@ const userType = new GraphQLObjectType({
...
+      description: 'the User’s search engine',
+      resolve: user => getUserSearchEngine(user),
+    },
+    showYahooPrompt: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      description: 'whether to show the yahoo search prompt',
+      resolve: user => !user.yahooSearchSwitchPrompt.hasSeenPrompt,
```
kmjennison, line 758: This should depend on the user's experiment group, right? Check experiment group; if the user's in the control or not assigned, false; else if in treatment,  true if they haven't already seen the prompt (false if they have already seen it).

graphql/database/users/getUserSearchEngine.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Fetch the user's search engine by the following procedure:
+ * 1. If set on the UserModel, return the value
+ * 2. If unset, see if a search widget value is set and migrate it, then return that value
+ * 3. If still unset, see if the user should be part of a test group to have a particular search engine and return that
+ * 4. If still unset, return the default search engine
+ * use exported logic for default search engine
+ * @param {object} userContext - The user authorizer object.
+ * @param {string} user - UserModel object.
+ * @return {Promise<User>}  A promise that resolves into a User instance.
```
kmjennison, line 17: Update to return a full SearchEngine object.

graphql/database/users/__tests__/getUserSearchEngine.test.js
```diff
@@ -0,0 +1,99 @@
...
+  it('defaults user engine to google if no widget, not in test or not set on model', async () => {
+    const getUserFeature = require('../../experiments/getUserFeature').default
+    getUserFeature.mockResolvedValue(undefined)
+    getWidgets.mockResolvedValueOnce([])
+
+    const result = await getUserSearchEngine(userContext, user)
+    expect(result).toEqual('Google')
+  })
+
+  it('defaults user engine to google if no widget, not in test or not set on model', async () => {
```
kmjennison, line 48: Update test description?

kmjennison (changes requested):
Directionally looking good.

graphql/data/schema.js
jedtan: Yeah that's fair. I was taking the ticket pretty literally and thinking we could combine it with the bucket in the front-end, but probably better to do the calculation in one place if we're able to do so.

web/data/schema.graphql
```diff
@@ -697,6 +697,15 @@ type failedEmailAddresses {
   error: String
 }
```
kmjennison, line 698: I think this needs to be rebuilt?

graphql/data/schema.js
```diff
@@ -746,6 +749,18 @@ const userType = new GraphQLObjectType({
       description: 'feature values for this specific user',
       resolve: (user, args, context) => getUserFeatures(context, user),
     },
```
kmjennison, line 751: Related: can you change the return type for setUserSearchEngineMutation? That way, the refetched data will update the client behavior appropriately.

graphql/data/schema.js
```diff
@@ -746,6 +749,18 @@ const userType = new GraphQLObjectType({
...
+      type: SearchEngineType,
+      description: 'the User’s search engine',
+      resolve: (user, args, context) => getUserSearchEngine(context, user),
+    },
+    showYahooPrompt: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      description: 'whether to show the yahoo search prompt',
+      resolve: user =>
+        !user.yahooSearchSwitchPrompt.hasSeenPrompt &&
+        getUserFeature(YAHOO_SEARCH_EXISTING_USERS).variation,
```
kmjennison, line 762: Please break this out into its own module. I know it's small, but we try to keep as much business logic out of schema.js as possible to make it more testable.

graphql/data/schema.js
kmjennison: (Sorry for not thinking to call that out last time.)

graphql/database/search/getSearchEngine.js
```diff
@@ -2,9 +2,12 @@ import { find } from 'lodash/collection'
 import searchEngines from './searchEngines'
 import { DatabaseItemDoesNotExistException } from '../../utils/exceptions'
 
-const getSearchEngine = async id => {
+const getSearchEngine = async (id, ignoreUnfound = false) => {
```
kmjennison, line 5: A better pattern is for the calling code to try/catch and specifically handle the DatabaseItemDoesNotExistException.

graphql/database/users/getUserSearchEngine.js
```diff
@@ -0,0 +1,57 @@
...
+  // Query Search Widgets and Find
+  const widgets = await getWidgets(userContext.user, user.id, false)
+  const maybeSearchWidget = widgets.filter(
+    widget => widget.type === WIDGET_TYPE_SEARCH
+  )
+  if (maybeSearchWidget.length > 0) {
+    const searchWidget = maybeSearchWidget[0]
+    const maybeSearchEngine = getSearchEngine(
+      JSON.parse(searchWidget.config).engine,
+      true
```
kmjennison, line 36: See other comment—use try/catch instead.

graphql/database/users/__tests__/getUserSearchEngine.test.js
```diff
@@ -0,0 +1,128 @@
...
+})
+
+beforeEach(() => {
+  jest.clearAllMocks()
+})
+const user = getMockUserInstance()
+
+describe('getUserSearchEngine', () => {
+  it('defaults user engine to google if no widget, not in test or not set on model', async () => {
+    const getUserFeature = require('../../experiments/getUserFeature').default
```
kmjennison, line 41: Add `expect.assertions(1)` to these async tests as an additional sanity check that assertions are called prior to the test's conclusion.

graphql/database/experiments/features.js
```diff
@@ -11,18 +16,32 @@ const features = {
...
+
+features[YAHOO_SEARCH_NEW_USERS] = {
+  defaultValue: 'Google',
+  rules: [
+    {
+      condition: {
+        isTabTeamMember: true,
+        env: 'local',
+      },
+      force: 'Yahoo',
```
kmjennison, line 42: Update to SFAC now that we're using that brand for charitable search.

graphql/database/users/__tests__/getUserSearchEngine.test.js
```diff
@@ -0,0 +1,128 @@
...
+    const result = await getUserSearchEngine(userContext, user)
+    expect(result).toEqual(await getSearchEngine('Google'))
+  })
+
+  it('defaults user engine to user feature value if no widget, or not set on model', async () => {
+    const getUserFeature = require('../../experiments/getUserFeature').default
+    getUserFeature.mockResolvedValue(
+      new Feature({
+        featureName: YAHOO_SEARCH_NEW_USERS,
+        variation: 'Google',
```
kmjennison, line 54: It would be a stronger test to use the non-Google option (SearchForACause) given that Google's the default here.

kmjennison (changes requested):
Good progress. Thanks for the iteration on this PR.

graphql/database/search/getSearchEngine.js
kmjennison: Example: https://github.com/gladly-team/tab/blob/587b2fc1f960924bb0e5871451dfb2164d4259cb/graphql/database/users/rewardReferringUser.js#L45

graphql/database/users/__tests__/getShouldShowYahooPrompt.test.js
```diff
@@ -0,0 +1,80 @@
...
+        featureName: YAHOO_SEARCH_EXISTING_USERS,
+        variation: true,
+      })
+    )
+
+    const result = await getShouldShowYahooPrompt(userContext, user)
+    expect(result).toEqual(true)
+  })
+
+  it('returns false if users feature is true for test', async () => {
```
kmjennison, line 66: Description could use an edit -- "false for test".

kmjennison (approved):
LGTM! 🦊

PR #1206, opened 2022-03-07
TFAC-894: Adding growthbook to tab and Setting users features on the …

jedtan:
…user type

### DIFF
--- a/graphql/data/schema.js
+++ b/graphql/data/schema.js
@@ -35,6 +35,7 @@ import {
   VIDEO_AD_LOG,
   USER_EXPERIMENT,
   SEARCH_ENGINE,
+  FEATURE,
 } from '../database/constants'
 
 import { experimentConfig } from '../utils/experiments'
@@ -119,6 +120,7 @@ import searchEngines from '../database/search/searchEngines'
 import UserExperimentModel from '../database/experiments/UserExperimentModel'
 import getSearchEngine from '../database/search/getSearchEngine'
 import SearchEngineModel from '../database/search/SearchEngineModel'
+import getUserFeatures from '../database/experiments/getUserFeatures'
 
 // eslint-disable-next-line import/no-named-as-default
 import getRecruits, {
@@ -738,8 +740,27 @@ const userType = new GraphQLObjectType({
       type: new GraphQLNonNull(GraphQLBoolean),
       description: 'whether a v4 user has been introduced to squads in the ui',
     },
+    userFeatures: {
+      type: new GraphQLList(featureType),
+      description: 'feature values for this specific user',
+      resolve: (user, args, context) => getUserFeatures(context, user),
+    },
   }),
   interfaces: [nodeInterface],
+})
+const featureType = new GraphQLObjectType({
+  name: FEATURE,
+  description: 'all important data a user assigned to a feature.',
+  fields: () => ({
+    featureName: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: `Name of the Search Engine`,
+    },
+    variation: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'the value of the variation for this specific user',
+    },
+  }),
 })
 const CauseImpactCopy = new GraphQLObjectType({
   name: 'CauseSpecificImpactUI',
--- a/graphql/data/schema.js
+++ b/graphql/data/schema.js
@@ -740,7 +740,7 @@ const userType = new GraphQLObjectType({
       type: new GraphQLNonNull(GraphQLBoolean),
       description: 'whether a v4 user has been introduced to squads in the ui',
     },
-    userFeatures: {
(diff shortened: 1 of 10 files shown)


kmjennison:
@jedtan This seems like it'll work. A few questions/thoughts:
1. One pitfall is that this returns all features data via API, even ones that are irrelevant or maybe not in use anymore. Would we just need to manage the list of features to keep it pruned? Do you know if GrowthBook automatically filters out inactive features (in the case we load features.json remotely, not defined inline)?
2. Is the tracking callback called for each feature every time this runs? I'm wondering if we'll have a problem with the process ending before all callbacks are fired. Should we `await` the callback / is that supported?
3. In this setup, it's fully possible for the frontend code to try to access a user-feature value for a feature that no longer exists. We'll need to handle that in frontent logic.
4. From a developer usability perspective on the frontend, do you see any issues with this?

jedtan:
1. Yes, we'd need to prune the list of features. I would expect us to actively prune experiments as part of ending the experiment process. The UI should also allow you to remove features should we host `features.json` elsewhere.
2. Yes, and you're right that this could be an issue. Thanks for bringing this to my attention. I asked the Growthbook folks and they mentioned that nothing is built in to await callbacks. They also mentioned that 
image
We could theoretically completely subvert the trackingCallback completely by just taking the values of the `.feature` call and seeing if the user was bucketed because of the experiment. I'm not really sure what the probability of this missed Mutation would be and whether or not the juice would be worth the squeeze.
3. Yes, this is also true. I would just consider this "feature not found" case the user not being bucketed in the experiment.
4. Asides from what you mention in #3, not really. We'll have to filter the list of features returned now, though I could reshape the returned object to be a map.

kmjennison:
Regarding #2, I like your suggestion to manage writing the DB values ourselves. It comes with the benefit of aggregating all items and batch-updating to the table. This all sounds good!

graphql/database/experiments/__tests__/getUserFeatures.test.js
```diff
@@ -0,0 +1,113 @@
...
+      new Feature({
+        featureName: 'test-feature',
+        variation: false,
+      }),
+      new Feature({
+        featureName: 'yahoo-search',
+        variation: 'variation-A',
+      }),
+    ])
+    expect(createUserExperiment).toHaveBeenCalledWith(
```
jedtan, line 88: Wondering if you could help me take a look at this unit test @kmjennison - not sure why this `expect` isn't working.

kmjennison: @jedtan You need to require the `createUserExperiment` module within the test. When you're using `jest.resetModules()` between tests, it's resetting _all_ modules and so the referenced imported module at the top of this test file won't be used.

jedtan: Ah, gotcha.

graphql/data/schema.js
```diff
@@ -738,9 +740,28 @@ const userType = new GraphQLObjectType({
...
+      type: new GraphQLList(featureType),
+      description: 'feature values for this specific user',
+      resolve: (user, args, context) => getUserFeatures(context, user),
+    },
   }),
   interfaces: [nodeInterface],
 })
+const featureType = new GraphQLObjectType({
+  name: FEATURE,
+  description: 'all important data a user assigned to a feature.',
```
kmjennison, line 753: Could be more descriptive

graphql/data/schema.js
```diff
@@ -738,9 +740,28 @@ const userType = new GraphQLObjectType({
...
+  fields: () => ({
+    featureName: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: `Name of the Search Engine`,
```
kmjennison, line 757: Edit description

graphql/data/schema.js
```diff
@@ -738,9 +740,28 @@ const userType = new GraphQLObjectType({
       type: new GraphQLNonNull(GraphQLBoolean),
       description: 'whether a v4 user has been introduced to squads in the ui',
     },
+    userFeatures: {
```
kmjennison, line 743: Naming feedback for the future: since this is a property on the user, you could just call this "features". The user connection is implied, I think.

graphql/database/experiments/__tests__/getUserFeatures.test.js
```diff
@@ -0,0 +1,96 @@
...
+      userContext,
+      user.id,
+      'yahoo-search',
+      'variation-A'
+    )
+  })
+})
```
kmjennison, line 96: Please add tests for `growthbook.setFeatures` and `growthbook.setAttributes` behavior.

kmjennison (changes requested):
A few small changes. @jedtan feel free to merge after addressing. Good work on this!

PR #1214, opened 2022-03-27
TFAC-891: Enabling SFAC, Yahoo as Search Engines

jedtan:
This will enable both engines. I also shoehorned some logic into the CreateSearchEnginePromptLogMutation to correctly update UserModel.


kmjennison (approved):
LGTM! ☕ Thanks for your diligence.

PR #1218, opened 2022-04-04
Modify LogSearch mutation to support global userID

kmjennison:
The LogSearch mutation currently expects a plain user ID. This PR supports passing the user's Relay global ID.


kmjennison:
@

PR #1221, opened 2022-04-06
Adding the yahooPaidSearchRewardOptIn Field to UserType

jedtan:


kmjennison (commented):
Linting fix needed, then LGTM

PR #1239, opened 2022-04-23
TFAC-806: New Search Engine Test and Experiment Definition Setup

jedtan:
This PR should not modify production behavior. It sets up the search experiment behavior on local and dev environments.


graphql/database/experiments/features.js
```diff
@@ -43,30 +43,60 @@ const features = {
...
-  //       env: 'local',
-  //     },
-  //     force: true,
-  //   },
-  // ],
+  rules: [
+    /* Begin internal overrides */
+    {
+      condition: {
+        v4BetaEnabled: true,
```
kmjennison, line 50: I think (but am not 100% confident) this has to use `$eq` instead because a simple `true` is truthy:
https://github.com/gladly-team/tab/pull/1235/files

graphql/database/experiments/features.js
```diff
@@ -43,30 +43,60 @@ const features = {
...
+          $eq: true,
+          $exists: true,
+        },
+      },
+      force: true,
+    },
+    {
+      condition: {
+        v4BetaEnabled: true,
+        'internalExperimentOverrides.search-existing-users-v2': {
```
kmjennison, line 61: Should this be the same key as the above? FYI, in case you didn't know, you can use variables in key names like:
```js
{
  [`my-string-and-${someVar}`]: "foo",
}
```

graphql/database/experiments/features.js
```diff
@@ -43,30 +43,60 @@ const features = {
...
+        'internalExperimentOverrides.search-existing-users': {
+          $eq: true,
+          $exists: true,
+        },
+      },
+      force: true,
+    },
+    {
+      condition: {
+        v4BetaEnabled: true,
```
kmjennison, line 60: Same as above.

graphql/database/experiments/features.js
```diff
@@ -43,30 +43,60 @@ const features = {
...
+    /* End internal overrides */
+    {
+      variations: [false, true],
+      weights: [0.5, 0.5],
+      coverage: 0.3,
+      condition: {
+        joined: {
+          $lt: 1650726528502, // make a later date when we go to production.
+        },
+        v4BetaEnabled: true,
```
kmjennison, line 77: Same as above.

graphql/database/experiments/features.js
```diff
@@ -43,30 +43,60 @@ const features = {
...
+    /* End internal overrides */
+    {
+      variations: ['Google', 'SearchForACause'],
+      weights: [0.5, 0.5],
+      coverage: 1.0,
+      condition: {
+        joined: {
+          $gt: 1650726528502, // make a later date when we go to production.
+        },
+        v4BetaEnabled: true,
```
kmjennison, line 96: Same as above.

graphql/database/users/createUser.js
```diff
@@ -164,6 +166,21 @@ const createUser = async (
...
   }
 
+  // Set user's search engine for new user
+  try {
+    const newSearchFeature = await getUserFeature(
+      userContext,
+      returnedUser,
+      YAHOO_SEARCH_NEW_USERS
+    )
+    await UserModel.update(userContext, {
```
kmjennison, line 176: Update `returnedUser` here so the returned value will be fully up-to-date.

graphql/database/users/__tests__/createUser.test.js
```diff
@@ -1000,4 +1028,36 @@ describe('createUser when user already exists (should be idempotent)', () => {
...
+    })
+
+    await createUser(userContext, userInfo.id, userInfo.email, null)
+
+    expect(updateMethod).toHaveBeenCalledWith(userContext, {
+      id: userInfo.id,
+      searchEngine: 'SearchForACause',
+      updated: moment.utc().toISOString(),
+    })
+  })
```
kmjennison, line 1062: Test returned user value has the correct search engine.

kmjennison (changes requested):

graphql/database/experiments/features.js
jedtan: If you click into that link, `v4BetaEnabled` in that condition is also `true`

jedtan: Is the truthy check just the case for that string key?

graphql/database/experiments/features.js
jedtan: Good call on the key being the same.

graphql/database/experiments/features.js
kmjennison: @jedtan Sorry, I was unclear. That PR link was just to show the use of `$eq: true` in the override attribute, which I'm suggesting we use anywhere we have a boolean match (even though the linked PR doesn't do so for v4BetaEnabled).

The behavior I observed was that setting an attribute value of `false` also matched on a simple `true` statement—so it's actually not "truthy" matching but whether or not the value is defined (?). I couldn't find clarity in the MongoDB docs, so this was just trial and error, and I'm not totally certain I'm right.

If I'm right, v4BetaEnabled matching wouldn't work at all (it would match any defined value). This doesn't matter right now because legacy doesn't request experiment info from the endpoint but would break in the future.

graphql/database/experiments/features.js
```diff
@@ -1,17 +1,21 @@
 import {
+  MONEY_RAISED_EXCLAMATION_POINT_V2,
   YAHOO_SEARCH_EXISTING_USERS,
   YAHOO_SEARCH_NEW_USERS,
 } from './experimentConstants'
 
+// TODO: update as needed before launch
+const EXPERIMENT_LAUNCH_UNIX_TIME = 1651075307324 // ~4pm UTC 27 April 2022
```
jedtan, line 8: nit, `SEARCH_EXPERIMENT_LAUNCH_UNIX_TIME`

graphql/database/users/getShouldShowYahooPrompt.js
```diff
@@ -15,9 +17,14 @@ const getShouldShowYahooPrompt = async (userContext, user) => {
...
+  const alreadyResponded =
+    user.yahooSearchSwitchPrompt &&
+    user.yahooSearchSwitchPrompt.hasRespondedToPrompt
+  const alreadyUsingSearchEngine =
+    user.searchEngine === charitableSearchEngineId
   return (
-    !user.yahooSwitchSearchPrompt &&
-    !user.yahooPaidSearchRewardOptIn &&
+    !alreadyResponded &&
+    !alreadyUsingSearchEngine &&
```
jedtan, line 27: How come there's no more `yahooPaidSearchRewardOptIn` being checked here?

kmjennison: @jedtan I figure what matters here is if the user's not using the charitable search engine. If they're **not** using the charitable search engine but previously opted in, we still want to ask them to switch search. Does my thinking make sense?

kmjennison: My bad: doing this broke the intended behavior for the new users experiment. Will revert.

kmjennison (approved):

PR #1255, opened 2022-05-14
TFAC-981: Creating Model.js and DynamoDBModel.js

jedtan:


graphql/database/base/Model.js
```diff
@@ -0,0 +1,469 @@
...
+      if (!isNil(val)) {
+        self[fieldName] = val
+      }
+    })
+  }
+
+  static register() {
+    // Kind of dumb logic for now, just makes sure all expected internal methods are defined.
+    /* eslint-disable no-unused-expressions */
+    this.schema
```
jedtan, line 74: Do I need this? This might be a little dumb.

kmjennison: This seems just fine to me. It's like build-time validation for whatever model extends this.

graphql/database/base/Model.js
```diff
@@ -0,0 +1,469 @@
...
+    this.hashKey
+    this.tableName
```
kmjennison, line 76: I don't think tableName is used by Model.

graphql/database/base/DynamoDBModel.js
```diff
@@ -0,0 +1,213 @@
...
+   *   operation name. Each property value is a function that receives:
+   *    - a userContext object
+   *    - item hashKey (if getting a specific item)
+   *    - item rangeKey (if one exists)
+   *    - item object (if creating or updating an item)
+   *   The authorizer function returns a boolean for whether the query is authorized.
+   *   Secondary indexes must be authorized separately in `indexPermissions`,
+   *   with a property key set to the name of the secondary index.
+   */
+  static get permissions() {
```
kmjennison, line 91: Seems like this isn't necessary here? It can just live on Model?

graphql/database/base/DynamoDBModel.js
```diff
@@ -0,0 +1,213 @@
...
+
+  static async getAllInternal() {
+    const items = await this.dynogelsModel
+      .scan()
+      .loadAll()
+      .execAsync()
+    return items.Items
+  }
+
+  static query(userContext, hashKey) {
```
kmjennison, line 141: Should this be here? I assume the "connector" should only implement the internal methods?

graphql/database/base/DynamoDBModel.js
```diff
@@ -0,0 +1,213 @@
...
+      getAll: () => false,
+      // Receives: userContext, hashKeyValue, rangeKeyValue, item
+      update: () => false,
+      // Receives: userContext, hashKeyValue, rangeKeyValue, item
+      create: () => false,
+      indexPermissions: {},
+    }
+  }
+
+  static async get(userContext, hashKey, rangeKey) {
```
kmjennison, line 105: Should this be here? I assume the "connector" should only implement the internal methods?

graphql/database/base/Model.js
kmjennison: If the expectation is to error to the developer, though, maybe check each of these and throw a more useful error message.

graphql/database/base/Model.js
```diff
@@ -0,0 +1,469 @@
...
+  static get name() {
+    throw new NotImplementedException()
+  }
+
+  /**
+   * The name of the database table.
+   * You are required to override this function on the child class.
+   * @return {string} The name of the database table.
+   */
+  static get tableName() {
```
kmjennison, line 95: Not needed in the base Model?

graphql/database/base/DynamoDBModel.js
jedtan: I kept it because of the `indexPermissions`. Open to doing it a different way.

graphql/database/base/DynamoDBModel.js
jedtan: The query returns a `queryObj` that additional filters can be applied to, which wouldn't be applicable to all `Model` instances.

graphql/database/base/DynamoDBModel.js
jedtan: Whoops, that's correct.

graphql/database/base/DynamoDBModel.js
kmjennison: Ok, that seems reasonable.

graphql/database/base/__tests__/Model-properties.test.js
```diff
@@ -0,0 +1,50 @@
...
+    }).toThrow('Model is missing a required field (schema, hashKey or name)')
+  })
+
+  it('fails if "schema" property is not set', () => {
+    const TestModel = require('../test-utils/ExampleModelV2').default
+    delete TestModel.schema
+    expect(() => {
+      TestModel.register()
+    }).toThrow('Model is missing a required field (schema, hashKey or name)')
+  })
```
kmjennison, line 40: Duplicate test

kmjennison (changes requested):
This looks great. I have one design change request: change the naming/pattern of the "internal" methods. Right now, they're named like "getInternal", which is a little confusing and would make it easier for a developer to accidentally call directly on an internal method rather than the intended public one. A couple of options to make this clearer:
1. Use the `_` naming convention, like `_get`, for the internal classes
2. Try marking methods as private, which I just saw [is making its way into the language](https://github.com/tc39/proposal-class-fields#private-fields)! I haven't used this before.
3. Override the methods in the child class but call `super` to first invoke the base model's logic.

graphql/database/base/__tests__/Model-properties.test.js
jedtan: For #2, we have to upgrade eslint and some other stuff. I couldn't get past it super easily but I'll work on it next.

For #3, I'm confused as to what you mean. Calling super is automatic if I don't have a constructor right? 
For methods, which would I call super in? For example, calling super in `_get` would throw a `NotImplementedException`. I also don't define the public methods in the subclasses.

PR #1257, opened 2022-05-22
Changing to DynamoDBModel

jedtan:
This branch excludes a few models:

- FeatureModel
- CauseModel
- SearchEngineModel

This is because these aren't based on a DynamoDB table anyways. The first one is just a data holder, and the latter two are based on json/js data (and will be converted as we eventually implement the JSONModel).


kmjennison (approved):
LGTM! ☕

PR #1259, opened 2022-05-26
TFAC-984 RedisModel Implementation

jedtan:
Some corners cut:
- Objects aren't supported as fields in RedisModel yet.
- There's not really a truly "atomic" `create` method.
- There's not really have a way to version RedisModels. If the schema changes, we may run into issues with fetching created objects.


graphql/database/base/__tests__/RedisModel-queries.test.js
```diff
@@ -0,0 +1,458 @@
...
+    setModelPermissions(ExampleRedisModel, {
+      create: () => true,
+      get: () => true,
+    })
+
+    return expect(
+      ExampleRedisModel.getField(user, 'dummy', 'name')
+    ).rejects.toEqual(new DatabaseItemDoesNotExistException())
+  })
+})
```
kmjennison, line 458: Nice work on these tests!

kmjennison (changes requested):
I am confused about how primary keys are defined for this model and how the keys interface with Model.js. There seem to be some inconsistencies, but it might just be that I need help understanding the approach, so let's talk live as needed.

graphql/database/base/RedisModel.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  static async wrapItem(dataPromise) {
+    return {
+      attrs: await dataPromise,
+    }
+  }
+
+  static getRedisKey(hashKey) {
+    return `${this.name}_${hashKey}`
```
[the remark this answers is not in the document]
jedtan: I think for the sake of simplicity we can just use a singular key for the model. I'm not sure what value a range key gets us since querying Redis for a range of keys is a super expensive operation.

graphql/database/base/RedisModel.js
```diff
@@ -0,0 +1,107 @@
...
+    await Promise.all(
+      Object.entries(item).map(([key, value]) =>
+        client.hset(redisKey, key, value)
+      )
+    )
+    return this.wrapItem(client.hgetall(redisKey))
+  }
+
+  static async updateInternal(item) {
+    const redisKey = this.getRedisKey(item.id)
```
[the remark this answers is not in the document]
kmjennison: @jedtan ^ I still have this question.

graphql/database/base/RedisModel.js
```diff
@@ -0,0 +1,107 @@
...
+    return this.wrapItem(item)
+  }
+
+  static async getBatchInternal(keys) {
+    const items = keys.map(key => this.getInternal(key))
+    return Promise.all(items)
+  }
+
+  static async createInternal(item, overwrite = false) {
+    const redisKey = this.getRedisKey(item.id)
```
kmjennison, line 28: @jedtan Same question as below: should the "id" property actually be the hash key?

graphql/database/base/RedisModel.js
```diff
@@ -0,0 +1,107 @@
...
+
+    const redisKey = this.getRedisKey(id)
+    const result = await client.hget(redisKey, field)
+
+    if (result === null) {
+      throw new DatabaseItemDoesNotExistException()
+    }
+
+    await client.hset(redisKey, field, value)
+    return client.hget(redisKey, field)
```
kmjennison, line 75: (Optimization; I have not looked at Redis docs) Are both the `hset` and `hget` necessary, or does `hset` return the value? Or, can you just return the `value` without getting it?

graphql/database/base/test-utils/ExampleRedisModel.js
```diff
@@ -0,0 +1,62 @@
+import types from '../../fieldTypes'
+import RedisModel from '../RedisModel'
+
+class ExampleRedisModel extends RedisModel {
+  static get name() {
+    return 'Example'
+  }
+
+  static get hashKey() {
+    return 'id'
```
kmjennison, line 10: Related to other comments, will this model still work with something other than `id` here?

graphql/database/base/RedisModel.js
jedtan: hset returns the # of fields created: https://redis.io/commands/hset/

We could always return the value though. I'll change that!

graphql/database/base/RedisModel.js
jedtan: Yep, that's correct.

graphql/database/base/RedisModel.js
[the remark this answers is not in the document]
jedtan: Yes, you're correct. We should be using the hash key value. I'll update that!

I'm not sure if the second comment (https://github.com/gladly-team/tab/pull/1259#discussion_r964952417) is still relevant after the last set of changes - let me know if that is consistent with what you see.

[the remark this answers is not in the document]
kmjennison: Yes, second comment is resolved after our discussion.

graphql/database/base/RedisModel.js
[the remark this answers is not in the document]
jedtan: Do you see an advantage for supporting a complex key over mostly treating the RedisModel closer to key/value store?

graphql/database/base/test-utils/ExampleRedisModel.js
jedtan: After the most recent updates, this model should still work with something other than id.

graphql/database/base/RedisModel.js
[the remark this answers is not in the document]
kmjennison: @jedtan I'm sorry, I apparently missed this notification. A composite key could be useful if we need to find an item based only on two other known attributes, such as "look up a user's impact from their user ID and a cause ID". With the current implementation, we're limited to creating unique identifiers and storing a reference to those IDs elsewhere. That works for now, and composite key behavior wouldn't be too hard to add if needed.

graphql/database/base/__tests__/RedisModel-authorization.test.js
```diff
@@ -0,0 +1,59 @@
...
+const user = getMockUserContext()
+
+describe('Model calls to `isQueryAuthorized`', () => {
+  afterEach(() => {
+    jest.resetModules()
+  })
+
+  it('passes correct params to `get` authorization check', async () => {
+    // Set a mock `isQueryAuthorized` method
+    const TestModelRangeKey = require('../test-utils/ExampleDynamoDBModelRangeKey')
```
kmjennison, line 13: This is using the wrong model. If you want to just drop this test file altogether, I'm fine with that.

kmjennison (dismissed):
Nice work! Tiny test change, then good to go.

kmjennison:
Regarding this limitation you mention in the PR description:

> There's not really have a way to version RedisModels. If the schema changes, we may run into issues with fetching created objects.

That's not a huge deal for us. We very rarely do DB migrations and typically rely on default fields in the model.

kmjennison:
Adding another limitation here for future reference: the key value doesn't support a composite key. A composite key could be useful if we need to find an item based only on two other known attributes, such as "look up a user's impact from their user ID and a cause ID". With the current implementation, we're limited to creating unique identifiers and storing a reference to those IDs elsewhere. That works for now, and composite key behavior wouldn't be too hard to add if needed.

PR #1261, opened 2022-05-31
TFAC-1034 Set up SFAC Experiment for New Users

jedtan:


graphql/database/experiments/features.js
```diff
@@ -135,6 +136,66 @@ const features = {
       },
     ],
   },
+  [YAHOO_SEARCH_NEW_USERS_V2]: {
+    defaultValue: 'Tooltip',
```
kmjennison, line 140: Default to control?

graphql/database/experiments/features.js
```diff
@@ -135,6 +136,66 @@ const features = {
...
+            $exists: true,
+          },
+        },
+        force: 'Notification',
+      },
+      /* End internal overrides */
+      {
+        variations: ['Control', 'Tooltip', 'Notification'],
+        weights: [0.34, 0.33, 0.33],
+        coverage: 1.0,
```
kmjennison, line 183: As-is, this would start assigning users to the experiment when deployed, right? Set coverage to zero.

graphql/database/experiments/features.js
```diff
@@ -135,6 +136,66 @@ const features = {
...
+        condition: {
+          isTabTeamMember: {
+            $eq: true,
+          },
+          joined: {
+            // todo, update gt time when launching. also end previous experiment
```
kmjennison, line 189: This should work to update both experiments at the same time. You could also set this experiment to exclude anybody in the previous experiment.

graphql/database/users/__tests__/createUser.test.js
```diff
@@ -1116,4 +1126,52 @@ describe('createUser when user already exists (should be idempotent)', () => {
 
     expect(returnedUser.searchEngine).toEqual('SearchForACause')
   })
+
```
kmjennison, line 1129: Seems like the v1 test is no longer testing v1 because it's using the same getUserFeature mock set up in `beforeEach` that has the v2 experiment enabled? If that's right, explicitly set the experiment mock there to keep the test to v1.

graphql/database/users/__tests__/createUser.test.js
```diff
@@ -1116,4 +1126,52 @@ describe('createUser when user already exists (should be idempotent)', () => {
...
+      userInfo.email,
+      null
+    )
+
+    expect(updateMethod).toHaveBeenCalledWith(userContext, {
+      id: userInfo.id,
+      searchEngine: 'SearchForACause',
+    })
+    expect(returnedUser.searchEngine).toEqual('SearchForACause')
+  })
```
kmjennison, line 1176: Please add tests for Notification and Control, too.

graphql/database/experiments/features.js
```diff
@@ -135,6 +136,66 @@ const features = {
...
+        condition: {
+          v4BetaEnabled: {
+            $eq: true,
+          },
+          [`internalExperimentOverrides.${YAHOO_SEARCH_NEW_USERS_V2}`]: {
+            $eq: 'Notification',
+            $exists: true,
+          },
+        },
+        force: 'Notification',
```
kmjennison, line 177: The notification behavior needs updating, such as in [getShouldShowYahooPrompt.js](https://github.com/gladly-team/tab/blob/92c8f855d0d5e2f6369a12bc8b7e41834e091e11/graphql/database/users/getShouldShowYahooPrompt.js), before it's functional.

kmjennison (changes requested):

graphql/database/experiments/features.js
jedtan: This should be 0, but just to clarify I intended the `isTabTeamMember` condition to gate that behavior.

graphql/database/experiments/features.js
jedtan: I'm not sure I follow this. Also, any reason why we can't merge without these dates being set if the coverage for v2 is 0?

jedtan: I feel like it'd be safe to just merge in both the date values when we actually turn on the coverage and remove the `isTabTeamMember` condition.

graphql/database/users/__tests__/createUser.test.js
jedtan: I think we still need to return something since `v2` exists, but that something should be correct now in the `beforeEach`.

graphql/database/experiments/features.js
kmjennison: That's my mistake. I overlooked that condition. 👍

graphql/database/experiments/features.js
kmjennison: That totally works. My comment about setting up an exclusion would be just an extra safeguard in this case and not necessary. We'd want it only if there's some reason to decouple ending the first experiment from launching the second.

graphql/database/users/__tests__/createUser.test.js
kmjennison: Yup, looks good 👍

kmjennison (approved):
LGTM! ☕ Nice work.

PR #1266, opened 2022-06-04
TFAC-1033: Start Experiment V2

jedtan:


graphql/database/experiments/features.js
```diff
@@ -186,8 +191,7 @@ const features = {
             $eq: true,
```
kmjennison, line 191: Need to remove

PR #1277, opened 2022-06-27
TFAC-1049: logSearch updates

jedtan:


graphql/database/users/logSearch.js
```diff
@@ -82,7 +84,55 @@ const logSearch = async (userContext, userId, searchData = {}) => {
...
+      isAnonymous: true,
+      version: 1,
+    })
+  } catch (e) {
+    throw e
+  }
+  return {
+    success: true,
+  }
+}
```
kmjennison, line 113: Might be able to make this more DRY with the overlap in logging/validation.

graphql/database/users/logSearch.js
```diff
@@ -82,7 +84,55 @@ const logSearch = async (userContext, userId, searchData = {}) => {
...
+/**
+ * Log a user's search event and change related stats.
+ * @param {object} userContext - The user authorizer object.
+ * @param {string} userId - The user id.
+ * @param {string} anonUserId - The anonymous user id.
+ * @return {Promise<User>} A promise that resolves into a User instance.
+ */
+const logSearch = async (userContext, userId, anonUserId, searchData = {}) => {
+  if (userId && anonUserId) {
+    throw new Error('userId and anonUserId cannot be set at once.')
```
kmjennison, line 124: Let's allow this case: prioritize using the userId if it exists, else use anonymous. (This is just one less error case to think about.)

graphql/database/users/logSearch.js
```diff
@@ -82,7 +84,55 @@ const logSearch = async (userContext, userId, searchData = {}) => {
...
+  }
+
+  if (userId) {
+    return logSearchKnownUser(userContext, userId, searchData)
+  }
+
+  if (anonUserId) {
```
kmjennison, line 131: If there's not an authenticated user, let's generate the anon user ID here and also set it in claims. This is for AC #4 on this ticket:
https://gladlyteam.atlassian.net/browse/TFAC-1048

Future-proofing: if the anonUserId exists, you can just use it and assume it has a claim set, too.

graphql/database/users/logSearch.js
```diff
@@ -62,18 +63,19 @@ const logSearch = async (userContext, userId, searchData = {}) => {
...
+        : (await getUserSearchEngine(userContext, user)).id
     const logPromise = UserSearchLogModel.create(userContext, {
       userId,
       timestamp: moment.utc().toISOString(),
       ...(source && { source }),
-      searchEngine: searchEngine.id,
+      searchEngine: searchEngineId,
       ...(causeId && { causeId }),
       isAnonymous: false,
       version: 1,
```
kmjennison, line 81: Version should take input.

graphql/database/users/logSearch.js
```diff
@@ -82,7 +84,55 @@ const logSearch = async (userContext, userId, searchData = {}) => {
...
+
+const logSearchAnonUser = async (userContext, anonUserId, searchData) => {
+  try {
+    const source = getSource(searchData)
+    await UserSearchLogModel.create(userContext, {
+      userId: anonUserId,
+      timestamp: moment.utc().toISOString(),
+      ...(source && { source }),
+      ...(searchData.searchEngineId && {
+        searchEngine: searchData.searchEngineId,
```
kmjennison, line 101: If no search engine is defined, default to SearchForACause.

kmjennison (changes requested):
Sorry—between our rearranging these two tickets ([TFAC-1048](https://gladlyteam.atlassian.net/browse/TFAC-1048), [TFAC-1049](https://gladlyteam.atlassian.net/browse/TFAC-1049))
 and cutting some features, I did not write complete ACs, and we lost some intended functionality.

Mainly:
* TFAC-1048 AC: A completely unauthed user—with no claims—can call /graphql/ to log to UserSearchLog.
* Set the anon ID per log. We won't send any anon user ID as input at this point. (From the planning doc.)
* TFAC-1048 AC: We “fake” the anonymous user claim (setting the claim to the nanoid when calling the model) to retain the model’s authorizer permissions without having to first implement additional authorizer logic. 
* The `version` should be an input that defaults to `1` but is optionally passed in the mutation.

web/data/schema.graphql
```diff
@@ -735,19 +735,27 @@ type LogReferralLinkClickPayload {
...
+  The actual user ID (not the Relay global). Provide either this or `userIdGlobal` for an existing user.
   """
   userId: String
+
+  """The anonymous user ID. Provide this for an anonymous user."""
+  anonUserId: String
   source: String
+  causeId: String
+  searchEngineId: String
   clientMutationId: String
```
kmjennison, line 753: Schema needs to be updated.

graphql/database/users/logSearch.js
```diff
@@ -1,18 +1,38 @@
...
+  version,
+  isAnonymous,
+  source
+) => {
+  return UserSearchLogModel.create(userContext, {
+    userId,
+    timestamp: moment.utc().toISOString(),
+    ...(source && { source }),
+    searchEngine: searchEngineId || 'SearchForACause',
+    ...(causeId && { causeId }),
```
kmjennison, line 29: This still needs some validation. See: https://github.com/gladly-team/tab/pull/1277#discussion_r907623912

plus some tests for logging null if the inputs don't pass validation.

kmjennison: + some tests for logging null if the inputs don't pass validation.

kmjennison (changes requested):
Looking good. Just a few edits.

kmjennison (approved):
LGTM! ☕ Please deploy to dev and do a manual QA prior to merge (or let me know if you want me to take that). Thanks for the iterations!

PR #1302, opened 2022-07-12
TFAC-1072 Call graphql with auth cookie values from subscribed search lambda

jedtan:


lambda/src/search-request-logger/search-request-logger.mjs
```diff
@@ -34,8 +35,13 @@ exports.handler = async event => {
...
       // https://docs.aws.amazon.com/apigateway/latest/developerguide/configure-api-gateway-lambda-authorization-with-console.html"
-      Authorization:
-        (message.user && message.user.idToken) || 'unauthenticated',
+      Authorization: (messageUser && messageUser.idToken) || 'unauthenticated',
+      ...(messageUser.authUserTokens && {
+        'X-Tab-Auth-User-Tokens': messageUser.authUserTokens,
+      }),
+      ...(messageUser.authUserTokensSig && {
+        'X-Tab-Auth-User-Tokens-Sig': messageUser.authUserTokensSig,
+      }),
```
kmjennison, line 44: Awaiting exploration of the requirements for the authorizer's identity sources before approving this. We may need to always send non-null values in the headers.

kmjennison: Confirmed that the authorizer requires all identity sources to be set, so we'll have to take a different approach. Discussing elsewhere.

lambda/src/search-request-logger/search-request-logger.mjs
```diff
@@ -1,6 +1,25 @@
...
 import fetch from 'node-fetch'
 
+const getAuthorizationHeaderFromMessage = messageUser => {
+  if (!messageUser) {
+    return 'unauthenticated'
+  }
+
+  if (messageUser.idToken) {
+    return messageUser.idToken
+  }
```
kmjennison, line 11: We can remove this. There's no case now where we'll pass an idToken value in the SNS message.

kmjennison (changes requested):
Removing idToken, otherwise looks great!

kmjennison (approved):
LGTM! ☕

PR #1303, opened 2022-07-13
TFAC-1073 Pass auth cookie values to SNS from edge search lambda

jedtan:


lambda/src/search-app-lambda-edge/search-app-lambda-edge.js
```diff
@@ -4,10 +4,54 @@
...
+import { parse } from 'cookie'
 import searchURLByRegion from './searchURLByRegion'
 
 const PRODUCTION_STAGE = 'prod'
 
+const decodeBase64 = string => {
+  return Buffer.from(string, 'base64').toString('utf8')
+}
+
+const getCookieVal = rawString => {
```
kmjennison, line 16: This is no longer needed. Rather than trying to decode or interpret the cookie values at all, we're just passing the raw values on and letting the authorizer do the work.

lambda/src/search-app-lambda-edge/search-app-lambda-edge.js
```diff
@@ -4,10 +4,54 @@
...
+// https://github.com/gladly-team/next-firebase-auth/issues/223
+const getIdTokensFromCookies = (cookiesStr = '') => {
+  const cookies = parse(cookiesStr)
+  console.log(cookies)
+
+  const authCookieName = 'TabAuth.AuthUserTokens'
+  const authCookieSigName = 'TabAuth.AuthUserTokens.sig'
+
+  const authCookieVal = cookies[authCookieName] || null
+  const authCookieSigVal = cookies[authCookieSigName] || null
```
kmjennison, line 47: This would get the job done. Remove logic below and rename function.

kmjennison (changes requested):

kmjennison (approved):
DNM yet, but LGTM! ☕ 

We need to wait to update the authorizer logic before merging this. Otherwise, the current authorizer will error because of the unexpected Authorization header value.

I'll handle merging when ready. Notes to self on doing so:
* Update the stack policy to allow changes to the web stack\
* Test on dev (after the updated authorizer is deployed there)
* After a deploy, set the new search lambda version number env var

PR #1306, opened 2022-07-21
TFAC-1067 Convert firebase-authorizer to async/await

jedtan:


lambda/src/firebase-authorizer/firebase-authorizer.js
```diff
@@ -73,79 +73,74 @@ function checkUserAuthorization(event, context, callback) {
...
+      if (err) {
+        console.log('Decrypt error:', err)
+        return 'Error decrypting secure environnment variables.'
+      }
+      decryptedFirebasePrivateKey = data.Plaintext.toString('ascii')
+      return checkUserAuthorization(event)
+    }
+  )
+
+  return new Error('Error: Unexpected end of function')
```
kmjennison, line 139: Because the KMS stuff above is callback-based, I think you'll be reaching this return any time the decrypted private key is not available.

lambda/src/firebase-authorizer/firebase-authorizer.js
```diff
@@ -73,79 +73,72 @@ function checkUserAuthorization(event, context, callback) {
...
+  } catch (err) {
+    console.log('Decrypt error:', err)
+    throw new Error('Error decrypting secure environnment variables.')
   }
 }
 
-const serverlessHandler = (event, context, callback) => {
-  handler(event, context, callback)
+const serverlessHandler = async event => {
+  handler(event)
```
kmjennison, line 141: You'll have to return the promise here.

kmjennison (changes requested):
I tried deploying to dev, and it failed with: `Error: [serverless-plugin-include-dependencies]: Could not find aws-crt`:
https://app.travis-ci.com/github/gladly-team/tab/builds/253810458

You might have to whitelist the new dependencies you want to include in the deployed package here:
https://github.com/gladly-team/tab/blob/master/lambda/serverless-lambda.yml#L90

An example of explicitly including dependencies in the build:
https://github.com/gladly-team/tab/blob/master/lambda/serverless-lambda-edge.yml#L95-L105

kmjennison (approved):
LGTM! ☕

PR #1311, opened 2022-08-06
TFAC-1092 Refactor authorizer decryption in prep for using next-firebase-auth

jedtan:


lambda/src/firebase-authorizer/firebase-authorizer.js
```diff
@@ -116,29 +117,26 @@ const checkUserAuthorization = async event => {
...
 }
 
+const _handler = async event => {
+  initNFA({
+    firebaseProjectId: process.env.LAMBDA_FIREBASE_PROJECT_ID,
+    firebasePrivateKey: decryptedFirebasePrivateKey.replace(/\\n/g, '\n'),
+    firebaseClientEmail: process.env.LAMBDA_FIREBASE_CLIENT_EMAIL,
+    firebaseDatabaseURL: process.env.LAMBDA_FIREBASE_DATABASE_URL,
+    firebasePublicAPIKey: process.env.FIREBASE_PUBLIC_API_KEY,
+    cookieKeys: [await decryptValue(process.env.COOKIE_SECRET_20220711)],
```
kmjennison, line 127: This will call `decryptValue` on each invocation, which probably isn't what you want?

lambda/src/firebase-authorizer/firebase-authorizer.js
```diff
@@ -116,29 +117,26 @@ const checkUserAuthorization = async event => {
   }
 }
 
+const _handler = async event => {
+  initNFA({
```
kmjennison, line 121: I think we should init only once (as needed) at the top level—but that's just an optimization, so ignore if it's annoying to implement.

lambda/src/firebase-authorizer/firebase-authorizer.js
```diff
@@ -116,29 +117,26 @@ const checkUserAuthorization = async event => {
...
+    firebaseDatabaseURL: process.env.LAMBDA_FIREBASE_DATABASE_URL,
+    firebasePublicAPIKey: process.env.FIREBASE_PUBLIC_API_KEY,
+    cookieKeys: [await decryptValue(process.env.COOKIE_SECRET_20220711)],
+  })
+  return checkUserAuthorization(event)
+}
+
 const handler = async event => {
   // Decrypt secure environment variables.
   if (decryptedFirebasePrivateKey) {
```
kmjennison, line 134: Suggestion (ignore as desired): you might add a single helper that verifies NFA has been initialized and that all necessary decrypted values are set, and if not, deals with doing those things.

lambda/src/firebase-authorizer/firebase-authorizer.js
jedtan: Oh, sounds good. I had thought we needed to init on each invocation.

kmjennison (commented):
Code looks good but this is 500ing on dev. I'm taking a look.

kmjennison (approved):
Tested on dev and LGTM! ☕

PR #1316, opened 2022-08-12
TFAC-1071: Support authorization via auth cookie values

kmjennison:


kmjennison:
I'm getting an unauthenticated user when testing on dev with what I believe is an Authorization header containing valid token values. Will debug.

lambda/src/firebase-authorizer/__tests__/firebase-authorizer.test.js
```diff
@@ -249,6 +290,136 @@ describe('firebase-authorizer', () => {
...
+          {
+            Action: 'execute-api:Invoke',
+            Effect: 'Allow',
+            Resource: 'arn:execute-api:blah:blah',
+          },
+        ],
+      },
+      context: {
+        id: 'abc123xyz987',
+        email: 'meow@hogwarts.com',
```
jedtan, line 340: nit, hogwarts.edu

lambda/src/firebase-authorizer/firebase-authorizer.js
```diff
@@ -77,8 +78,30 @@ const checkUserAuthorization = async event => {
...
+    })
+    const user = {
+      uid: nfaUser.id,
+      email: nfaUser.email,
+      email_verified: nfaUser.emailVerified,
+      auth_time: 0, // NFA does not provide an auth time
+    }
+    return generatePolicy(user, true, event.methodArn)
+  } catch (e) {
+    // If not, don't throw. This is expected when provided with a Firebase
```
jedtan, line 100: Are there any other non - JSON parse errors we could get there? Might be useful to log if it's not an expected error type.

jedtan:
Largely looks good to me, interested in hearing what the issue is.

lambda/src/firebase-authorizer/firebase-authorizer.js
kmjennison: Good suggestion, will update.

kmjennison:
It looks like the cookie secret is set or encrypted incorrectly. Will fix and re-test.

PR #1317, opened 2022-08-13
TFAC-1090 Updated UserModel

jedtan:
First change, just to keep track on a per user basis who has seen the SFAC Extension prompt.


kmjennison (approved):
LGTM ☕

PR #1318, opened 2022-08-16
TFAC-1090 Backend Presentational Logic

jedtan:
Won't be used yet since the FE isn't hooked up. Also I will add browser logic for Safari later once I figure out the best way to do that!


graphql/database/experiments/features.js
```diff
@@ -198,6 +199,47 @@ const features = {
...
+          },
+        },
+        force: 'Control',
+      },
+      {
+        condition: {
+          v4BetaEnabled: {
+            $eq: true,
+          },
+          [`internalExperimentOverrides.${YAHOO_SEARCH_NEW_USERS_V2}`]: {
```
kmjennison, line 223: Fix name of override

graphql/database/experiments/features.js
```diff
@@ -198,6 +199,47 @@ const features = {
       },
     ],
   },
+  [SFAC_EXTENSION_PROMPT]: {
+    defaultValue: 'Control',
```
kmjennison, line 203: Though this won't yet be used, users would be bucketed into the experiment if we deploy this PR: calling `user.features` in schema will get all features and bucket the user. This would mess with analytics.

Does that sound right to you? If so, comment out the actual experiment here until we're ready to launch.

graphql/database/experiments/features.js
```diff
@@ -198,6 +199,47 @@ const features = {
...
+    rules: [
+      /* Begin internal overrides */
+      {
+        condition: {
+          v4BetaEnabled: {
+            $eq: true,
```
kmjennison, line 209: Add the same "tabs" condition to both of these overrides so our internal testing reflects actual behavior.

graphql/database/experiments/features.js
```diff
@@ -198,6 +199,47 @@ const features = {
...
+      },
+      /* End internal overrides */
+      {
+        variations: ['Control', 'Notification'],
+        weights: [0.5, 0.5],
+        coverage: 0.0,
+        condition: {
+          tabs: {
+            $gt: 3,
+          },
```
kmjennison, line 238: Add "v4BetaEnabled" for good measure. That's just future-proofing, in case we start fetching these features on Legacy.

graphql/database/users/__tests__/getShouldShowSFACExtensionPrompt.test.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    const result = await getShouldShowSfacExtensionPrompt(
+      userContext,
+      modifiedUser
+    )
+    expect(result).toEqual(false)
+  })
+
+  it('returns true if the user is in the experiment and has responded', async () => {
```
kmjennison, line 38: Edit: "has not responded"

kmjennison (changes requested):

graphql/database/experiments/features.js
jedtan: I thought that it wouldn't if there's no coverage yet. No one does get bucketed in that case no?

kmjennison: @jedtan That's my bad, I missed the coverage property 🤦 This should be all good.

kmjennison (approved):
LGTM! ☕

PR #1319, opened 2022-08-16
TFAC-1114 Create new UserEventLog Table

jedtan:


graphql/database/logs/logTypes.js
```diff
@@ -0,0 +1,24 @@
+import types from '../fieldTypes'
+
```
kmjennison, line 2: The typing here isn't strictly necessary but is probably helpful as a guide to others consuming data and as one additional check on data soundness. I'm on board with it.

graphql/database/logs/UserEventLogModel.js
```diff
@@ -0,0 +1,49 @@
...
+import tableNames from '../tables'
+import { permissionAuthorizers } from '../../utils/authorization-helpers'
+import { VALID_LOG_TYPES } from './logTypes'
+
+/*
+ * @extends DynamoDBModel
+ */
+class UserEventLogModel extends DynamoDBModel {
+  static get name() {
+    return ''
```
kmjennison, line 12: Will need a table name here.

graphql/database/logs/logTypes.js
jedtan: Yeah I think the former was from my perspective as it serves as documentation in one spot for the differing log schemas.

graphql/database/logs/UserEventLogModel.js
```diff
@@ -0,0 +1,52 @@
...
+/*
+ * @extends DynamoDBModel
+ */
+class UserEventLogModel extends DynamoDBModel {
+  static get name() {
+    return USER_EVENT_LOG
+  }
+
+  static get hashKey() {
+    return 'userId'
```
jedtan, line 17: @kmjennison Would like your thoughts on this. If we happen to want to write two logs in the same mutation technically the hashKey/rangeKey could collide.

jedtan: I feel like something ideal would be event type / user id / timestamp; is there a way to express this that way?

kmjennison: That said, I've elsewhere cut corners on the userId+timestamp index problem by assuming some small number of item creations will conflict and just making micro-adjustments to the timestamps:
https://github.com/gladly-team/tab/blob/master/graphql/database/users/logTab.js#L59-L63

kmjennison (commented):
Awaiting your opinion/edits on table design.

graphql/database/logs/UserEventLogModel.js
jedtan: That's definitely a creative way to do it!

kmjennison (approved):
LGTM! ☕

PR #1329, opened 2022-09-02
TFAC-985 Implementing ImpactMetric Model

jedtan:


graphql/database/groupImpact/ImpactMetricModel.js
```diff
@@ -0,0 +1,65 @@
...
+          `the dollar amount required to achieve an instance of this ImpactMetric`
+        ),
+      description: types
+        .number()
+        .required()
+        .description(`Description of the ImpactMetric`),
+      name: types
+        .number()
+        .required()
+        .description(`Name of the impact metric`),
```
kmjennison, line 46: It's unclear what the "name" and "description" fields are used for. If you have clarity about their purpose, please improve the descriptions. If you're also unsure, let's just drop the fields for now and spec later what fields we need based on the UI.

In mocks, I see an impact metric having a few things:
- A kind of metric title: "1 home visit"
- An impact title: "Provide 1 visit from a community health worker"
- A description:
 > Living in the communities in which they work, community health workers are trusted neighbors who know their community best and use their linguistic, cultural, and technical expertise.

> This provides access to care for people who might not otherwise have it.

I don't think we ever finalized spec for those fields. Related templating ticket:
https://gladlyteam.atlassian.net/browse/TFAC-957

graphql/database/groupImpact/ImpactMetricModel.js
```diff
@@ -0,0 +1,65 @@
...
+class ImpactMetric extends DynamoDBModel {
+  static get name() {
+    return IMPACT_METRIC
+  }
+
+  static get hashKey() {
+    return 'id'
+  }
+
+  static get tableName() {
```
kmjennison, line 17: Just confirming: we'll be drawing from JSON/YML data for the near term, right? No need for this to live in a DB right now.

kmjennison (changes requested):

graphql/database/groupImpact/ImpactMetricModel.js
jedtan: I think that I'll throw in those three fields that you've mentioned into the model for now. We can also bring in the ticket you link into the current sprint so we can start looking at the spec doc. I'll do another pass of the draft too just to get it ready for a look over!

kmjennison: Sounds great!

kmjennison (approved):
LGTM! ☕

PR #1339, opened 2022-09-20
TFAC-985 GroupImpactMetric Model

jedtan:


graphql/database/groupImpact/GroupImpactMetricModel.js
```diff
@@ -0,0 +1,72 @@
...
+class GroupImpactMetric extends RedisModel {
+  static get name() {
+    return GROUP_IMPACT_METRIC
+  }
+
+  static get hashKey() {
+    return 'id'
+  }
+
+  static get tableName() {
```
kmjennison, line 19: Nit: I think this isn't required?

kmjennison (changes requested):
Small edit, then LGTM!

PR #1346, opened 2022-09-26
Adding new ImpactMetric and GroupImpactMetric Types to schema.js

jedtan:
I did run `updateSchema`, but I presume there were no changes to schema.graphql since no APIs or return types changed.

### DIFF
--- a/graphql/data/schema.js
+++ b/graphql/data/schema.js
@@ -38,6 +38,8 @@ import {
   SEARCH_ENGINE_PERSONALIZED,
   FEATURE,
   CAUSE_IMPACT_TYPES,
+  IMPACT_METRIC,
+  GROUP_IMPACT_METRIC,
 } from '../database/constants'
 
 import { experimentConfig } from '../utils/experiments'
@@ -144,6 +146,7 @@ import getShouldShowSfacExtensionPrompt from '../database/users/getShouldShowSfa
 import getUserNotifications from '../database/users/getUserNotifications'
 import getSfacActivityState from '../database/users/getSfacActivityState'
 import getShouldShowSfacIcon from '../database/users/getShouldShowSfacIcon'
+import { getImpactMetricById } from '../database/groupImpact/impactMetricRepository'
 
 class App {
   constructor(id) {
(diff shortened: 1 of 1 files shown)


graphql/data/schema.js
```diff
@@ -965,6 +968,87 @@ const CauseOnboardingCopyType = new GraphQLObjectType({
...
+
+const impactMetricType = new GraphQLObjectType({
+  name: 'ImpactMetric',
+  description: 'An instance of GroupImpact',
+  fields: () => ({
+    id: {
+      type: globalIdField(IMPACT_METRIC),
+      description: 'ID of the ImpactMetric',
+    },
+    charityId: {
```
kmjennison, line 980: Use Charity type directly here. (Rule of thumb, if referencing the ID of another object, just use the object, which allows the client to query on additional fields from that object.)

graphql/data/schema.js
```diff
@@ -965,6 +968,87 @@ const CauseOnboardingCopyType = new GraphQLObjectType({
...
+// eslint-disable-next-line no-unused-vars
+const groupImpactMetric = new GraphQLObjectType({
+  name: 'GroupImpactMetric',
+  description: 'A specific instance of GroupImpactMetric',
+  fields: () => ({
+    id: {
+      type: globalIdField(GROUP_IMPACT_METRIC),
+      description: 'The ID of the GroupImpactMetric',
+    },
+    causeId: {
```
kmjennison, line 1020: Use Cause type directly here.

kmjennison (changes requested):

graphql/data/schema.js
jedtan: I initially didn't want to do that as these entities would be returned as a field of the Charity type. Wouldn't that potentially cause some sort of circular dependency?

kmjennison: I don't see how it would cause a problem. The resolver won't be called unless the field is queried, so I think the only way it would be kind-of circular is if someone made an intentionally deep nested query. If someone did query `charity > impact metric > charity` (I'd hope we'd catch this in code review), it should just resolve.

On the point of "would be returned as a field of the Charity type" -- this is true for what we can predict now, but I don't think the types need to make assumptions about their parent entities.

LMK if you feel I'm missing something!

kmjennison (commented):
Awaiting feedback/edits on the entity/resolver discussion.

graphql/data/schema.js
jedtan: Ah, thank you for the explanation! That totally makes sense. Thanks.

PR #1352, opened 2022-10-01
Replace ioredis with Mock

jedtan:


graphql/__mocks__/ioredis.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-env jest */
+const ioredisMock = jest.requireActual('ioredis-mock')
```
kmjennison, line 2: I'm a little confused by the purpose of this. The `requireActual` is getting the underlying module and then exporting it again, which should have no effect. Do you mean to create a mock from the module?

kmjennison (changes requested):

graphql/__mocks__/ioredis.js
jedtan: From how I understand having this as `ioredis.js` means that the ioredis package is being mocked, and I'm basically replacing it's full implementation with `ioredis-mock` instead.

kmjennison: Forgive my poor skimming here. I managed to miss the `-mock`! 🙄

kmjennison (commented):
LGTM! ☕

graphql/__mocks__/ioredis.js
jedtan: No worries!

PR #1353, opened 2022-10-02
TFAC-1137 Adjusting backend for notification for Legacy Users

jedtan:


kmjennison (approved):
LGTM ☕ Good catch here!

PR #1366, opened 2022-10-18
TFAC-1011 Create Feature Flag for Global Health Group Impact

jedtan:


graphql/utils/feature-flags.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/prefer-default-export */
+
+export const isGlobalHealthGroupImpactEnabled = () =>
```
kmjennison, line 3: Is the rationale for not using Growthbook that we currently require user info at runtime to initialize Growthbook? If so, we could create a user-agnostic feature-getter (`getFeatureAtBuildTime` or `getFeatureWithoutUser`).

Your approach works here, and no need to change it, but maybe worth adding a comment in this file that it's a one-off situation because it gets a little weird having two feature flag systems.

graphql/database/cause/__tests__/CauseModel.test.js
```diff
@@ -200,6 +200,7 @@ describe('CauseModel', () => {
     })
   })
 
+  // todo: @jtan fix test
```
kmjennison, line 203: Remove TODO?

kmjennison (approved):
LGTM ☕

graphql/utils/feature-flags.js
jedtan: Nope! I think it makes sense to keep it consistent. I'll switch it to a growthbook feature. For some reason I just interpreted "feature flag" very literally and implemented it similarly to how we implemented other feature flags.

jedtan: Here's a branch with my initial attempts. `jtan/tfac-1011-2`

kmjennison: Sounds good! If we edit this in the future, it likely makes sense to use async even for models fetching in-memory data.

PR #1370, opened 2022-10-28
TFAC-1157 Initial Experiment Setup for 3rd Ad on v4

jedtan:
Per Growthbook, they say that passing in the `time since joined` attribute should be done as a `setAttribute` call. They don't implement the mongo db matchers that would let us calculate it dynamically.


graphql/database/experiments/features.js
```diff
@@ -308,6 +310,27 @@ const features = {
...
+      {
+        // variations: [false, true],
+        // weights: [0.1, 0.9],
+        // coverage: 1.0,
+        condition: {
+          v4BetaEnabled: {
+            $eq: true,
+          },
+          joined: {
+            $lt: V4_SHOW_THIRD_AD_CUTOFF_TIME,
```
kmjennison, line 326: We want to include users who have joined >3 weeks ago rather than who joined after a specific datetime. This might require adding a field to the Growthbook user attributes to capture "time since joined" unless there's a MongoDB-style matcher that gets the job done here.

graphql/database/experiments/features.js
```diff
@@ -308,6 +310,27 @@ const features = {
       },
     ],
   },
+  [V4_SHOW_THIRD_AD]: {
+    defaultValue: false,
+    rules: [
+      // Show on local/dev for our team only.
```
kmjennison, line 316: Comment looks inaccurate.

kmjennison (changes requested):

graphql/database/experiments/features.js
kmjennison: Sorry, yes, your edit is what I meant to say.

kmjennison (approved):
LGTM! ☕ Thanks for exploring the Growthbook support for this.

PR #1372, opened 2022-11-02
TFAC-1158 local-user-data-mgr Helpers for Logged out Notification

jedtan:


web/src/js/utils/local-user-data-mgr.js
```diff
@@ -357,3 +359,30 @@ export const setUserDismissedSurvey2022 = () => {
 export const hasUserDismissedSurvey2022 = () => {
   return localStorageMgr.getItem('tab.user.dismissedUserSurvey2022') === 'true'
 }
+
+export const getLoggedOutMessage = () => {
+  const installTime = getBrowserExtensionInstallTime()
+
+  if (!installTime) {
+    return LOGGED_OUT_MESSAGE_TYPE.NONE
+  }
```
kmjennison, line 368: Let's edit this logic (sorry for the unclear spec): consider a user "new" if there is an install time and it's recent; otherwise, assume the user is old. It's unlikely an existing user who's logged out will have stuff set in local storage, so the current logic would probably rarely show to existing users.

kmjennison (changes requested):

web/src/js/utils/__tests__/local-user-data-mgr.test.js
```diff
@@ -525,4 +529,57 @@ describe('local user data manager', () => {
...
+    localStorageMgr.setItem(
+      STORAGE_APPROX_EXTENSION_INSTALL_TIME,
+      now.utc().toISOString()
+    )
+    localStorageMgr.setItem(STORAGE_LOGGED_OUT_TABS, 7)
+    const { getLoggedOutMessage } = require('js/utils/local-user-data-mgr')
+    expect(getLoggedOutMessage()).toBe(LOGGED_OUT_MESSAGE_TYPE.OLD)
+  })
+
+  it('returns old user logged out message if install date recent', () => {
```
kmjennison, line 561: "old" should be "new"

kmjennison (approved):
Small typo in test but LGTM! ☕

Issue #573, opened 2019-05-17
Background settings: custom image and color props should not be required


kmjennison:
Instead, set defaults.

PR #1387, opened 2022-11-28
Add a default backgroundColor so app doesn't error

pamplemoussecache:
This PR Is meant to partially address issue #573 . It only addresses the backgroundColor, not the backgroundImage error, as I wasn't sure what placeholder you would want to use (perhaps the `brokenImage`?) or if you would want to do any refactoring of this section of the codebase from classes to hooks.

### DIFF
--- a/web/src/js/components/Background/BackgroundColorPickerComponent.js
+++ b/web/src/js/components/Background/BackgroundColorPickerComponent.js
@@ -13,6 +13,7 @@ class BackgroundColorPicker extends React.Component {
 
     this.state = {
       selectedColor: '#000',
+      backgroundColor: '#000000'
     }
   }
 


kmjennison:
@pamplemoussecache Thanks for jumping in with a PR! Appreciate it!

I'm going to close this because:
* The component here does have a fallback color if the user's color isn't defined ([source](https://github.com/gladly-team/tab/blob/0c7a29d1edf8a75ead6d092f717fb41319fa7cfe/web/src/js/components/Background/BackgroundColorPickerComponent.js#L22))
* My original issue was unfortunately not very detailed, so I'm not certain what I was referring to—but I think we've handled the underlying error on the new tab page
* The primary focus of our frontend development is happening here: https://github.com/gladly-team/tab-web

Our product roadmap, along with bugs/specs, lives outside of GitHub, so I apologize that this issue was not up to date. It's a reminder to me to close the old issues here.

pamplemoussecache:
Gotcha! Thanks for the additional info, I'll go check out tab-web!

PR #1409, opened 2022-12-15
TFAC-978: modify group impact in dev environments

kmjennison:
* On local and dev, reduce the cost of impact metrics to $0.01 for simplified testing
* Replace custom feature-flags logic with Growthbook for enabling group impact on dev for the Global Health cause


web/data/schema.graphql
```diff
@@ -772,19 +772,12 @@ type ImpactMetric implements Node {
   """Charity ID that this impact metric belongs to"""
   charity: Charity
 
-  """
-  Dollar amount (in micro USDs) required to achieve an instance of this ImpactMetric
-  """
+  """Dollar amount required to achieve an instance of this ImpactMetric"""
```
jedtan, line 775: Should this change be here?

graphql/database/experiments/__tests__/getFeature.test.js
```diff
@@ -0,0 +1,42 @@
...
+import Feature from '../FeatureModel'
+import {
+  getConfiguredGrowthbook,
+  getFeatureWithoutUser,
+} from '../growthbookUtils'
+
+jest.mock('../growthbookUtils')
+
+describe('getFeature', () => {
+  it('configures Growthbook expected', async () => {
```
jedtan, line 11: nit: "expected Growthbook" or "Growthbook as expected"

graphql/database/experiments/growthbookUtils.js
```diff
@@ -7,9 +7,9 @@ import features from './features'
 import logger from '../../utils/logger'
 import { GROWTHBOOK_ENV } from '../../config'
 
-const validateAttributesObject = (userId, attributes) => {
-  // Could always use joi or similar if needed later.
-  const requiredProperties = [
+const validateAttributesObject = (userId, attributes, { noUserAttributes }) => {
+  const alwaysRequiredProperties = []
```
jedtan, line 11: Good forwards thinking!

graphql/database/experiments/growthbookUtils.js
```diff
@@ -35,6 +38,7 @@ export const getConfiguredGrowthbook = ({
   email,
   internalExperimentOverrides = {},
   tabs,
+  noUserAttributes = false,
```
jedtan, line 41: nit: I think the name `noUserAttributes` is a bit confusing. Could we find a way to avoid double negatives and change the param name to like... `userAgnostic` or `userSpecific` (the value for `userSpecific` being default true)?

graphql/database/experiments/__tests__/growthbookUtils.test.js
```diff
@@ -113,6 +113,14 @@ describe('growthbookUtils tests', () => {
     )
   })
 
+  test('getConfiguredGrowthbook does not warn on missing user attributes when the attributes are expected', async () => {
```
jedtan, line 116: are not*

jedtan (commented):
Left some comments after a first pass!

graphql/database/experiments/growthbookUtils.js
kmjennison: Thanks! Changed.

web/data/schema.graphql
kmjennison: Nope, reverted. Good catch, thanks!

PR #1411, opened 2022-12-16
Fix Upstash connection error

kmjennison:


jedtan (approved):
LGTM! Thanks for making these changes.

PR #1425, opened 2023-01-11
TFAC-1205 Shift to @upstash/redis Client

jedtan:


graphql/database/base/RedisModel.js
```diff
@@ -179,9 +164,8 @@ class RedisModel extends Model {
       const validateResult = this.schema[field].validate(fieldValue, {
         convert: true,
       })
-      if (validateResult.error) {
-        throw validateResult.error
-      }
+      // todo: @jedtan Implement Solution here for RedisModels
```
jedtan, line 167: This is because the `convert: true` parameter doesn't convert numbers -> strings properly. I don't think this is going to cause any immediate issues except for IDs occasionally being numbers in memory instead of strings (if a nanoid/uuid happens to be numeric by chance).

graphql/database/base/__tests__/RedisModel-queries.test.js
```diff
@@ -150,13 +158,13 @@ describe('RedisModel queries', () => {
     expect(createdItem).toEqual(expectedReturn)
   })
 
-  it('correctly creates item with default fields', async () => {
+  it.only('correctly creates item with default fields', async () => {
```
kmjennison, line 161: Remove `.only`

kmjennison (approved):
Small edit to a Jest test file, then good to go!

PR #1436, opened 2023-01-24
TFAC-1208: Good Enough Mock Redis

jedtan:
This will replace actual upstash which should be good for test timeouts.


kmjennison (approved):
LGTM! ☕ Thank you for the fix!

PR #1437, opened 2023-01-25
Installed js for getadmiral.com

spicermatthews:
Installs the js software snippet for getadmiral.com in the head of the index.html


spicermatthews:
Seems simple enough. Self approving. cc/ @kmjennison @jedtan

PR #1445, opened 2023-02-02
Removed Jan2023 notification and added Feb2023

spicermatthews:
This pull requests removes the Jan 2023 SFAC notification and adds the new Feb2023 notification.

In Jan we voted on which cause to support. This notification announces the winner.

### DIFF
--- a/graphql/database/users/getUserNotifications.js
+++ b/graphql/database/users/getUserNotifications.js
@@ -1,6 +1,6 @@
 import logger from '../../utils/logger'
 import getUserFeature from '../experiments/getUserFeature'
-import { NOTIF_SFAC_JAN_2023 } from '../experiments/experimentConstants'
+import { NOTIF_SFAC_FEB_2023 } from '../experiments/experimentConstants'
 
 /**
  * Get data for notifications the user should see.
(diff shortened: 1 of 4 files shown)


graphql/database/users/getUserNotifications.js
```diff
@@ -16,12 +16,12 @@ const getUserNotifications = async (userContext, user) => {
...
       userContext,
       user,
-      NOTIF_SFAC_JAN_2023
+      NOTIF_SFAC_FEB_2023
     )
     const enabled = notifFeature.variation !== 'None'
     notifications = [
       ...(enabled
-        ? [{ code: NOTIF_SFAC_JAN_2023, variation: notifFeature.variation }]
+        ? [{ code: NOTIF_SFAC_FEB_2023, variation: notifFeature.variation }]
```
kmjennison, line 24: The variation won't be used in this notification because it's not an experiment. However, I realize I made a mistake when defining [the `variation` field in the GraphQL schema](https://github.com/gladly-team/tab/blob/master/graphql/data/schema.js#L710): I set it to non-null when it should be nullable.

In summary: you can leave it for now, but at some point change the the schema to allow that field to be null.

kmjennison (approved):
LGTM!

PR #1453, opened 2023-02-16
TFAC-1015 Starting global health Group Impact

jedtan:


graphql/database/cause/causes.js
```diff
@@ -31,7 +31,7 @@ const overrideCauseModel = (cause, index) => {
     ).variation
     causes[index].impactType = globalHealthGroupImpactEnabled
       ? CAUSE_IMPACT_TYPES.group
-      : CAUSE_IMPACT_TYPES.none
+      : CAUSE_IMPACT_TYPES.individual
```
jedtan, line 34: This should also fix the bug here, though it won't matter at first since we're turning this feature on to true.

https://gladly-team.slack.com/archives/C03QBS615/p1675896768911439

spicermatthews (approved):
LGTM

PR #1459, opened 2023-03-01
Removing About The Text from Cause Data

jedtan:
This is going to be built into the component, so removing the "About The" text for each cause.


spicermatthews (approved):
LGTM

PR #1462, opened 2023-03-07
Build oAuth for SHFC

spicermatthews:
The wild fire extension needs an oAuth flow so we can return a token to the extension.

The Graphql goal for this PR is to build a query wildfire can call and get the name of a cause (and likely some additional information in the future) based on a user id. What makes his challenging is wildfire does not have an access token. Just a user id. So we added a new root called `wildfire` to our graphql schema. You can get information from this root without being authed.

```
query {
	wildfire(userId: "USERIDHERE") {
		causeName
	}
}
```

We use our context overrides to get around the model auth'ing the request.

On the front end we are building a flow. The idea is the shop for a cause extension will will redirect to our auth with something like this `/newtab/auth/?uri=chrome-extension%3A%2F%2Fextensionidhere%2Fauth.html`. Once done with the auth process we redirect back to the `uri` with the user's user id. This is how wildfire knows who the user is.


graphql/database/cause/__tests__/getCauseForWildfire.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+    const UserModel = require('../../users/UserModel').default
+    jest.spyOn(UserModel, 'get').mockResolvedValue(mockUser)
+
+    const getCauseForWildfire = require('../getCauseForWildfire').default
+    const cause = await getCauseForWildfire(userId)
+    expect(cause).toEqual(MOCK_CAUSE_2)
+  })
+
+  it('throws if the user id does not exist from wildfire', async () => {
```
jedtan, line 43: Can we add a test for if there is no causeId on the user?

web/src/js/components/Dashboard/DashboardComponent.js
```diff
@@ -441,7 +441,7 @@ class Dashboard extends React.Component {
 
     const isYahooUser = user && user.id === YAHOO_USER_ID
 
-    const notificationsToShow = this.state.notificationsToShow
+    // const notificationsToShow = this.state.notificationsToShow
```
jedtan, line 444: Looks like this might be an accident?

jedtan (commented):
Just a few comments, but overall largely looking good!

web/src/js/components/Dashboard/DashboardComponent.js
spicermatthews: @jedtan This should have been commented out before. It is an unused var at the moment. My editor gave me a warning.

spicermatthews:
@jedtan Thanks! I am going to merge as we need this now. Will add test in another PR.

PR #1541, opened 2023-06-17
Expose more Dollar Contribution Fields

jedtan:


spicermatthews (approved):
Maybe this is an opportunity to refactor a bit?

In an effort to clean up this file a bit I created a types directory. 

https://github.com/gladly-team/tab/blob/master/graphql/data/types/wildfire.js

Maybe we can move userGroupImpactMetricType into that directory?

If you are focused on other things don't worry. Just a suggestion.

jedtan:
That's a good idea! Will move.

